PR #75, opened 2025-04-04
s2/docker compose

alukach:
## What I'm changing

This builds out our docker-compose to support the services required to run the application.  This includes:

* NextJS
* DynamoDB
* Ory Tunnel

It also includes a DynamoDB bootstrap service that will populate the DynamoDB instance if no tables are discovered. 

## How I did it

* The DynamoDB instance is persisted to local disk, meaning that data will persist across docker compose runs.
* The `scripts/init-local.ts` script was altered to make it idempotent, only deleting existing tables if a `RESET_TABLES` env variable is set (currently commented-out in `docker-compose.yaml`)
* Getting a working Ory Tunnel was a bit of a challenge. The tunnel needs a valid auth session from Ory. To get such a session, we have two options:
  
  1. Expect user to authenticate locally (`ory auth login`). This will spin up a local webserver and take you through the auth flow, eventually writing your session details to `~/.ory-cloud.json`. We can then mount the session to the docker compose service and use that file to connect to Ory from within Docker. The downside is that it a) expects that you have the `ory` CLI on your machine, b) makes assumption about where your session is saved (I'm not sure that `~/.ory-cloud.json` will work on windows)
  2. Authenticate with ory from within the docker service.  The way that this works is that Ory creates a webserver with a random high port and [uses that as a redirect URL](https://github.com/ory/cli/blob/4cb1ec98bc269f5d7c50f32bdc49583179efa9f1/cmd/cloudx/client/auth.go#L183-L187) when performing the auth flow with Ory. This presents a challenge in that we can't know which port to expose from Docker Compose to the host machine ahead of time. We could attempt to expose all high ports, but that introduces a likely risk of port collision with the host machine and will cause Docker Compose to fail when attempting to run the service. Alternatively, we could run Ory Tunnel with `network_mode: host` which binds it to the host machine's network.
      ```yaml
        ory-tunnel:
          image: oryd/ory:latest
          network_mode: host
      ``` 
  The downside is that it requires a one-time manual setup step on the host machine (https://docs.docker.com/engine/network/drivers/host/#docker-desktop).
  
  I went with option 1).
* Rename `ORY_ACCESS_TOKEN` env var to `ORY_PROJECT_API_KEY` to match the env var expected by `ory tunnel`
* Small bugfix where I believe NextJS was using the wrong Ory client to communicate with the service from the backend (02f80be863728887fdc7776c95efba5b2bcadcf6)


jedsundwall (commented):
This approach sounds good. I'll see if I can get this to run locally.

jedsundwall (approved):

PR #76, opened 2025-04-04
S2: fix builds

alukach:
A militant rework of the codebase to get `npm run build` to pass.

### DIFF
--- a/examples/storage/local-example.ts
+++ b/examples/storage/local-example.ts
@@ -1,5 +1,5 @@
 import { LocalStorageClient } from '../../src/lib/storage';
-import { StorageProvider, StorageProviderConfig } from '../../src/types/storage';
+import { StorageProvider, StorageConfig } from '../../src/types/storage';
 
 async function main() {
   // Set up provider and config
@@ -11,7 +11,7 @@ async function main() {
     updated_at: new Date().toISOString(),
   };
 
-  const config: StorageProviderConfig = {
+  const config: StorageConfig = {
     type: 'LOCAL',
     endpoint: './data/storage',
   };
@@ -29,7 +29,11 @@ async function main() {
   try {
     // Store a file
     const content = Buffer.from('Hello, World!');
-    const objectInfo = await storage.putObject(testPath, content);
+    const objectInfo = await storage.putObject({
+      ...testPath,
+      data: content,
+      contentType: 'text/plain'
+    });
     console.log('Stored object:', objectInfo);
 
     // Read it back
--- a/scripts/cleanup-test-accounts.ts
+++ b/scripts/cleanup-test-accounts.ts
@@ -11,7 +11,7 @@ async function cleanupTestAccounts() {
   // Get list of test accounts
   const response = await fetch('/api/accounts', {
     headers: {
-      'x-csrf-token': session.csrf_token
+      // 'x-csrf-token': session.csrf_token
     },
     credentials: 'include'
   });
@@ -26,7 +26,7 @@ async function cleanupTestAccounts() {
       await fetch(`/api/accounts/${account.account_id}`, {
         method: 'DELETE',
         headers: {
-          'x-csrf-token': session.csrf_token
+          // 'x-csrf-token': session.csrf_token
         },
         credentials: 'include'
       });
--- a/src/app/[account_id]/[repository_id]/[...path]/page.tsx
+++ b/src/app/[account_id]/[repository_id]/[...path]/page.tsx
@@ -17,7 +17,6 @@ import { notFound } from 'next/navigation';
 import { ObjectBrowser, RepositoryHeader } from '@/components/features/repositories';
 
 // Types
-import type { Repository_v2 } from '@/types/repository_v2';
 import type { RepositoryObject } from '@/types/repository_object';
 
 // Utilities
@@ -25,11 +24,11 @@ import { fetchRepository } from '@/lib/db/operations_v2';
 import { createStorageClient } from '@/lib/clients/storage';
 
 interface PageProps {
-  params: {
+  params: Promise<{
     account_id: string;
     repository_id: string;
     path?: string[];
-  };
+  }>;
 }
 
 export default async function RepositoryPathPage({
--- a/src/app/api/accounts/route.ts
+++ b/src/app/api/accounts/route.ts
@@ -8,7 +8,7 @@ export async function POST(request: Request) {
     
     // Verify the user is authenticated
     const session = await ory.toSession();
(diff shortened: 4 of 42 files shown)


examples/storage/local-example.ts
```diff
@@ -29,7 +29,11 @@ async function main() {
   try {
     // Store a file
     const content = Buffer.from('Hello, World!');
-    const objectInfo = await storage.putObject(testPath, content);
+    const objectInfo = await storage.putObject({
+      ...testPath,
+      data: content,
+      contentType: 'text/plain'
+    });
```
alukach, line 36: Not sure that this will actually work.

scripts/cleanup-test-accounts.ts
```diff
@@ -11,7 +11,7 @@ async function cleanupTestAccounts() {
   // Get list of test accounts
   const response = await fetch('/api/accounts', {
     headers: {
-      'x-csrf-token': session.csrf_token
+      // 'x-csrf-token': session.csrf_token
```
alukach, line 14: Couldn't find CSRF token in session, so simply ignoring. This may likely break this code (although, APIs rarely require CSRF tokens in the header in my experience).

src/app/[account_id]/[repository_id]/[...path]/page.tsx
```diff
@@ -17,19 +17,18 @@ import { notFound } from 'next/navigation';
...
 import { createStorageClient } from '@/lib/clients/storage';
 
 interface PageProps {
-  params: {
+  params: Promise<{
     account_id: string;
     repository_id: string;
     path?: string[];
-  };
+  }>;
```
Comment (alukach) [inline src/app/[account_id]/[repository_id]/[...path]/page.tsx:31]: Apparently, all route params are async: https://nextjs.org/docs/app/building-your-application/routing/route-handlers#dynamic-route-segments

jedsundwall (approved):
😍

src/app/api/accounts/route.ts
```diff
@@ -8,23 +8,23 @@ export async function POST(request: Request) {
...
       return NextResponse.json(
         { error: 'No identity found in session' },
         { status: 400 }
       );
     }
 
-    const userId = session.identity.id;
-    const email = session.identity.traits.email;
+    const userId = session.data.identity.id;
+    const email = session.data.identity.traits.email;
```
alukach, line 27: This falls into my "how did this ever work?" bucket.

src/app/api/auth/logout/route.test.ts
```diff
@@ -1,6 +1,5 @@
 import { POST } from './route';
 import { cookies } from 'next/headers';
-import { _NextResponse } from 'next/server';
```
alukach, line 3: There's a lot of this `import { _Somevar } ...` convention in this codebase. Placing an underscore in front of the variable renames the variable and breaks the import.  I think commenting the line out is a better strategy moving forward.

src/app/layout.tsx
```diff
@@ -31,8 +31,10 @@ export default function RootLayout({
           storageKey="source-theme"
         >
           <SessionProvider>
-            <Box style={{ minHeight: '100vh' }}>
-              <Navigation />
+            <Box style={{ minHeight: "100vh" }}>
+              <Suspense>
+                <Navigation />
+              </Suspense>
```
alukach, line 37: Any SSR component (ie a page) will need to wrap any component that has `useAuth()` in `<Suspense>` (https://nextjs.org/docs/messages/prerender-error)

src/components/features/repositories/RepositorySchemaMetadata.tsx
```diff
@@ -29,7 +29,7 @@ export function RepositorySchemaMetadata({
       "creator": {
         "@type": account.type === 'organization' ? "Organization" : "Person",
         "name": account.name,
-        ...(account.website && { "url": account.website })
+        ...(account.websites && { "url": account.websites[0] })
```
alukach, line 32: @jedsundwall Wasn't sure about this move, but it was needed to make types happy.

tests/fixtures/example-data.ts
```diff
@@ -10,7 +10,8 @@ const exampleAccounts = {
     updated_at: '2024-01-01T00:00:00Z',
     owner_account_id: 'user1',
     admin_account_ids: ['user1'],
-    website: 'https://radiant.earth'
+    website: 'https://radiant.earth',
+    ory_id: 'radiant-earth'
```
alukach, line 14: Added a dummy `ory_id` to each of these

alukach:
@jedsundwall I see that you've approved, but could you take a quick 👀 at the comments I just left.  Those were the parts of this that stood out to me.

src/components/features/repositories/RepositorySchemaMetadata.tsx
jedsundwall: This looks right to me. We will be storing an array of websites associated with each account whereas we previously only allowed one.

src/app/layout.tsx
jedsundwall: Is this true for *any* SSR component? Certainly we have a lot of content that should be displayed regardless of authentication status.

src/app/api/accounts/route.ts
jedsundwall: 🪄

src/app/[account_id]/[repository_id]/[...path]/page.tsx
Comment (jedsundwall) [reply src/app/[account_id]/[repository_id]/[...path]/page.tsx:31]: I feel like they shouldn't be async because the app is constructed to always know the account_id and repository_id based on the URL. If the URL doesn't correspond to actual IDs, we should just provide a 404.

examples/storage/local-example.ts
jedsundwall: I think we can probably start getting rid of everything related to "local storage." I used that to build out the UI, but now we should only be working on data that's hosted in object stores.

jedsundwall:
Just responded to everything. I'm still busy with calls today, but will be available on Slack intermittently if you have more questions!

examples/storage/local-example.ts
alukach: Dropped in fabe917

src/app/[account_id]/[repository_id]/[...path]/page.tsx
Comment (alukach) [reply src/app/[account_id]/[repository_id]/[...path]/page.tsx:31]: While I agree with you on principle, NextJS seems to have made them async in version >14[^1]

[^1]: https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes#good-to-know

PR #77, opened 2025-04-04
Rm unreferenced test fixtures

alukach:
These test fixtures are declared but never referenced.  Similar fixtures exist in `src/tests/fixtures/example-data.ts` and are referenced by tests.  Recommend dropping these.


jedsundwall (approved):
Yes agree. This is old test data. Clean it out!

PR #78, opened 2025-04-04
Centralize all process.env reads into src/lib/config

alukach:
For the sake of understanding what env vars drive this system, I recommend mv all reads of `process.env` into `src/lib/config.ts`


jedsundwall (approved):
Agree with the move to `src/lib/config.ts`

PR #80, opened 2025-04-09
Productization

jedsundwall:
I've gone through the code base and replaced all mentions of "Repositories" with "Products." Moving forward, we will refer to "data products" rather than "repositories." The rationale for this is that we want our users to publish data in such a way that considers its potential application and users rather than a mere repository of content. More of our thoughts on this are explained in this blog post: https://radiant.earth/blog/2023/05/we-dont-talk-about-open-data/

The app loads properly with this branch, but it restarts whenever I try to load http://localhost:3000/clay or any product owned by `clay`.


alukach:
@jedsundwall seeing the following error on Vercel:

```
./scripts/convert-dynamo-data.ts:7:70
Type error: Cannot find module '../src/types/repository_v2' or its corresponding type declarations.
   5 | import { join } from 'path';
   6 | import type { Account, IndividualAccount, OrganizationalAccount } from '../src/types/account_v2';
>  7 | import type { Repository_v2, RepositoryMirror, RepositoryRole } from '../src/types/repository_v2';
     |                                                                      ^
   8 |
   9 | // Utility function to extract value from DynamoDB attribute format
  10 | function extractValue(attr: any): any {
Next.js build worker exited with code: 1 and signal: null
```

Not sure what happened to `../src/types/repository_v2`...

PR #82, opened 2025-04-11
Fix prod DynamoDB and S3 access

alukach:


jedsundwall (approved):
Looks good to me. I'd like to walk through s3.ts with you at some point to get an explanation of what it's doing and why (like, why do we use `systemPrefix`?), but just for my own education. Ship it!

src/lib/storage/s3.ts
```diff
@@ -141,9 +136,9 @@ export class S3StorageClient implements StorageClient {
 
   async getObjectInfo(params: GetObjectParams): Promise<RepositoryObject> {
     try {
-      const command = new GetObjectCommand({
-        Bucket: this.provider.endpoint,
-        Key: `${params.account_id}/${params.repository_id}/${params.object_path}`,
+      const command = new HeadObjectCommand({
```
alukach, line 139: Slight efficiency upgrade

PR #106, opened 2025-08-22
Add heading ID generation for MarkdownViewer component

### LINKED ISSUE #104: Headings within markdown files are no longer rendered as anchors
Once upon a time, named anchors within a markdown file made their way into the HTML on SC, and GitHub-style named anchors were generated for headings, so you could, e.g., create a TOC within a README file, like you might on GitHub.  For example, in [this file](https://beta.source.coop/repositories/agentmorris/lila-wildlife/description), I have a link like this:

`[North American Camera Trap Images](#north-american-camera-trap-images)`

That used to link to this heading, which became a named anchor in the rendered HTML:

`### North American Camera Trap Images`

It would be great to either (a) get that functionality back or (b) know that it was removed deliberately, in which case I can (I think) insert named anchors directly in the markdown (e.g. `<a name="north-american-camera-trap-images"></a>`, so long as those are propagated through to the HTML.

Not urgent at all, and FWIW despite this minor hiccup, the new rendering looks amazing.

Thanks!

alukach:
## What I'm changing

Implement a utility function to create URL-friendly IDs from heading text. Update heading components in MarkdownViewer to use this function, enhancing accessibility and linking capabilities.

## How I did it

We read the header text, massage into a URL-friendly and expect form, set that as the id of the header.

## How to test it

See the preview deployment: https://source-cooperative-git-fix-104ids-in-markdo-57f7df-radiantearth.vercel.app/agentmorris/lila-wildlife

**Related Issue(s):**

Closes #104

### DIFF
--- a/src/components/features/markdown/MarkdownViewer.tsx
+++ b/src/components/features/markdown/MarkdownViewer.tsx
@@ -17,6 +17,16 @@ interface MarkdownViewerProps {
   content: string;
 }
 
+// Utility function to convert heading text to URL-friendly ID
+function generateHeadingId(text: string): string {
+  return text
+    .toLowerCase()
+    .replace(/[^\w\s-]/g, '') // Remove special characters except spaces and hyphens
+    .replace(/\s+/g, '-') // Replace spaces with hyphens
+    .replace(/-+/g, '-') // Replace multiple hyphens with single hyphen
+    .trim();
+}
+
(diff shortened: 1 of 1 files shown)


alukach:
> we implemented a Markdown rendering React component but I'm not sure if we're still using it
I think that's what we're using now.  I have done a bit of work on it to make images work but I think this seems like a reasonable system to go forward with.

PR #112, opened 2025-08-26
Add working Create Organization form

### LINKED ISSUE #113: [Bug] Unable to create organizations
Currently, the Organization creation is disabled in production.  While the form exists, nothing occurs when a user submits the form.

Image

alukach:
**Related Issue(s):** #

Closes #113

**Proposed Changes:**

1. This PR adds some foundational work in generating forms with [server actions](https://nextjs.org/docs/app/getting-started/updating-data) to better integrate with the NextJS ecosystem. This will reduce the amount of APIs needed to support the frontend.
2. There were no flags set on an account when they were transferred over to the new DynamoDB tables. I have updated the script to embrace a set of defaults for each user. These should be confirmed & the migrations on production should be rerun once this PR is merged.  https://github.com/source-cooperative/source.coop/blob/c1e5a77b59bb3af416a452908bb1b3947adc784e/src/types/shared.ts#L31-L36


src/types/shared.ts
```diff
@@ -26,6 +28,13 @@ export enum AccountFlags {
   CREATE_ORGANIZATIONS = "create_organizations",
 }
 
+export const DEFAULT_INDIVIDUAL_FLAGS: AccountFlags[] = [
+  AccountFlags.CREATE_REPOSITORIES,
+  AccountFlags.CREATE_ORGANIZATIONS,
+];
```
alukach, line 34:
```suggestion
export const DEFAULT_INDIVIDUAL_FLAGS: AccountFlags[] = [];
```

@jedsundwall Confirming that we don't want users to be able to create orgs or repos by default?

jedsundwall: Yes this is correct. We need to grant them permission.

jedsundwall (approved):

PR #119, opened 2025-08-27
adding download urls

jedsundwall:
Adding URLs to download individual objects, both within the object browser and on the individual object page.

### DIFF
--- a/src/components/features/products/object-browser/DirectoryList.tsx
+++ b/src/components/features/products/object-browser/DirectoryList.tsx
@@ -1,10 +1,11 @@
 "use client";
 
-import { Box, Text, Flex } from "@radix-ui/themes";
+import { Box, Text, Flex, Button, Grid } from "@radix-ui/themes";
 import {
   ChevronRightIcon,
   FileIcon,
   ChevronDownIcon,
+  DownloadIcon,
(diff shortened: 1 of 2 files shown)


alukach:
Odd spacing on the list view:

image

Should we not just make the Source URL a link?

image

src/components/features/products/object-browser/DirectoryList.tsx
```diff
@@ -70,20 +71,24 @@ function DirectoryRow({
...
+        <Box gridColumn="1 / 8">
+          <Link
+            href={href}
+            className={styles.item}
+            title={item.name}
+            ref={(el: HTMLAnchorElement | null) => {
+              if (el && itemRefs) itemRefs.current[index] = el;
+            }}
+            onFocus={() => setFocusedIndex?.(index)}
+            data-focused={focusedIndex === index}
```
alukach, line 89: This feels like a lot of work for something that could be achieved via the [`:hover` CSS psuedo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover)

jedsundwall:
@alukach I agree with you on everything. Hopefully this looks good to you!

PR #123, opened 2025-09-03
Display error when we fail to list S3 object

alukach:
Currently, when the data proxy returns an error, the UI hides the issue and states that there are no files.  This is particularly common when running the project locally, wherein the data proxy doesn't know of a product that is in the dev fixtures:

image

This PR reworks the logic to display an error to acknowledge the failure:

image


jedsundwall (approved):
This looks fine, but I'd like us to have more specific error messages if we can. Like, shouldn't we be able to tell if a repository/bucket exists but that a specific object within it doesn't exist? I'm approving this, but I might create an issue to create more specific 404 messages in the future.

Issue #72, opened 2025-02-19
Add other viewers from components


severo:
Currently only the MapViewer is used.

We can use:
- the text viewer
- the table viewer (for CSV)
- the markdown viewer (already handled for README, let's merge some code)

Youssef-Harby:
I can work on that ... implement duckdb-wasm to quickly visualize CSV, Prquet .. other supported formats of duckdb spatial and wasm
It will work on client side, no backend for now

severo:
Good ideas.

The scope of this issue is narrower for two reasons:
1. it aims at using on source.coop the four viewer components already developed in https://github.com/source-cooperative/components/tree/main/packages/components (showcased here: https://source-cooperative.github.io/components/viewers/ and used here: https://viewers.source.coop/)
2. beyond this issue, our goal on source.coop is to show a quick preview of the contents of any file, without many features. That's why, for example, we don't plan to rely on duckdb-wasm but on hyparquet (and maybe hightable) for parquet files. See https://github.com/source-cooperative/components/issues/45. Adding viewers or improving the existing ones will be done for now at https://github.com/source-cooperative/components. Feel free to review the issues there if you want to start contributing!

jedsundwall:
@severo curious to get your opinion on @bdon's approach in #125. I discussed this with @alukach today and he and I agree that there's something elegant about having the viewer be entirely separate and shown via an iframe. This particularly makes sense in the case of PMTiles which benefits from having a sophisticated viewer maintained by its owner. It might not make as much sense for a simple CSV viewer. I defer to you on the approach you'd like to take for Parquet, GeoParquet, or other viewers you'd like to work on.

One nice thing about @bdon's approach is that pmtiles.io is simply a GitHub page, hosted entirely for free.

bdon:
I'd also propose building a COG Viewer for Source that lives inside an iframe:

* replace https://cogeo.org/map/ (relies on tiles.rdnt.io lambda)
* use https://github.com/geomatico/maplibre-cog-protocol so the user experience / basemap is similar to the PMTiles viewer

I think this new viewer should be in either the https://github.com/cogeotiff or http://github.com/source-cooperative GitHub orgs. 

Potential complexity:

* non-RGB COGs - how to default visualize?
* Reject GeoTIFFs that are not COGs?
* Overlap with existing viewers at http://labs.geomatico.es/maplibre-cog-protocol/ and https://geotiffjs.github.io/cog-explorer/ ?

[...]

PR #125, opened 2025-09-04
Embed a PMTiles preview via iframe to pmtiles.io

bdon:
**Related Issue(s):** #72 

This embeds [pmtiles.io](https://pmtiles.io) as an interactive preview on the object browser, when a file with the suffix `.pmtiles` is being viewed. 

**Proposed Changes:**

1. Add a new Card based on the suffix (we should make this more extensible to other data types - `.geojson`, `.tiff` etc)
2. For `pmtiles` add an `iframe` element that points to the `source.coop` public URL. These URLs are configured to allow embedding with CORS, so it won't work on all possible pmtiles URLs. 

Screenshot:

CleanShot 2025-09-04 at 16 31 15@2x

**PR Checklist:**

- [ ] This PR is made against the dev branch (all proposed changes except releases should be against dev, not main).
- [ ] This PR has **no** breaking changes.
- [ ] I have updated or added new tests to cover the changes in this PR.
- [ ] I have updated the version in `package.json` to follow the [Semantic Versioning](https://semver.org/) guidelines.
- [ ] All tests are passing.
- [ ] I have added my changes to the [CHANGELOG](https://github.com/source-cooperative/source.coop/blob/dev/CHANGELOG.rst)
      **or** a CHANGELOG entry is not required.
- [ ] This PR affects the [Source Cooperative Data Proxy](https://github.com/source-cooperative/data.source.coop),
      and I have opened issue/PR #XXX to track the change.


alukach:
@jedsundwall Sure, although this PR is still in "draft" mode. I recommend we wait for @bdon to confirm that it's good to go before merging

bdon:
> @bdon do you feel like this is ready to move from draft status to being merged?

Yes, it's fine in this state although we should add some more structure later around dealing with different extensions.

alukach:
> > @bdon do you feel like this is ready to move from draft status to being merged?
> 
> Yes, it's fine in this state although we should add some more structure later around dealing with different extensions.

Agreed.  I spent some time yesterday breaking up the `ObjectBrowser` component (see #151), creating an [`ObjectPreview`](https://github.com/source-cooperative/source.coop/pull/151/files#diff-2b469e3063d0704a671f9dff92c90213a042f3796587edd24b42e12129807272) where I think we can centralize the translation between extensions (or possible MIME types in the future) to viewers.

bdon:
I changed the styling on pmtiles.io in iframe mode to match the Source colors and layout better (screenshots):

CleanShot 2025-09-19 at 14 31 32@2x
CleanShot 2025-09-19 at 14 31 44@2x

PR #137, opened 2025-09-11
Revert "Display error when we fail to list S3 object"

alukach:
Reverts source-cooperative/source.coop#123

While it initially seemed to make to render errors on lookup failures, this is now seeming to be a not-good idea. Namely, for the reason that when a Product is first created, it is an empty "bucket" (ie path prefix in the Source S3 bucket). As such, it appears to be non-existent and the error is rendered.  However, this is a totally normal state simply representing an empty product/bucket-prefix and should not be rendered as an error.


alukach:
I believe that this was resolved in #139

It's possible that a better examination of errors returned by the proxy should take place in the future, but that is not yet possible due to https://github.com/source-cooperative/data.source.coop/issues/88

PR #139, opened 2025-09-12
Build a status page component for error views

alukach:
## What I'm changing

In this PR, we add common components for Forbidden and NotFound errors to make it easier to have unified error views across the application.

## How I did it

* Buildout `src/components/core/StatusPage.tsx` to handle NotFound and NotAuthorized error pages
* Update `not-found.tsx` files to use new `NotFoundPage` component

## How you can test this

Review the appropriate error pages:

* https://source-cooperative-git-refactor-build-out-e-f9db9b-radiantearth.vercel.app/foo
* https://source-cooperative-git-refactor-build-out-e-f9db9b-radiantearth.vercel.app/foo/bar
* https://source-cooperative-git-refactor-build-out-e-f9db9b-radiantearth.vercel.app/foo/bar/asdf/xyz

PR #141, opened 2025-09-16
Fix missing product error message and navigation optimizations

alukach:
## What I'm changing

In the process of #139, I noticed an error where a not-found product page would render a not-found account error:
  image
  

## How I did it

Fixing this took longer than expected and required a reorg of the product view, utilizing a [route group](https://nextjs.org/docs/app/api-reference/file-conventions/route-groups).  I attempted to document this here:
  https://github.com/source-cooperative/source.coop/blob/4ef482ec480c4000bf9420f4a12c680df0d7c6e1/src/app/%5Baccount_id%5D/%5Bproduct_id%5D/(product)/%5B%5B...path%5D%5D/layout.tsx#L1-L9

  While under the hood, I also made a slight optimization in the object viewing where the breadcrumbs now render instantly, no longer waiting until a directory of objects is loaded:

  https://github.com/user-attachments/assets/3feab4f1-6e7b-4230-a8c3-1f2a88308047

## How you can test it

Review the following error pages:

* https://source-cooperative-git-refactor-product-vie-5eebcf-radiantearth.vercel.app/foo/bar
* https://source-cooperative-git-refactor-product-vie-5eebcf-radiantearth.vercel.app/foo/bar/asdf/xyz

They should now show a Product Not Missing view instead of an Account Not Missing view.

Navigate through a product's files, only values that are changing should show loading skeleton.

### DIFF
--- a/src/components/features/products/object-browser/ObjectBrowser.tsx
+++ b/src/components/features/products/object-browser/ObjectBrowser.tsx
@@ -1,8 +1,10 @@
+import { Card } from "@radix-ui/themes";
+import { SectionHeader } from "@/components";
+import { DataConnection, Product, ProductMirror, ProductObject } from "@/types";
+
 import { ObjectDetails } from "./ObjectDetails";
 import { DirectoryList } from "./DirectoryList";
 import "./ObjectBrowser.module.css";
-import { ProductObject } from "@/types/product_object";
-import { DataConnection, Product, ProductMirror } from "@/types";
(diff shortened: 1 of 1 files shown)


alukach:
> I'd like to eventually harmonize how our error pages look a bit

Well, you're in luck because that is the intention #139!

> or modify the known-owner-but-unfound-product error to include some of the "chrome" that relates to the product owner so users can one-click back to see the other products published by the owners.

Let's take this convo to that PR.

PR #148, opened 2025-09-17
fix: set default mirrors on newly created products

alukach:
## What I'm changing

When a new product was created via the form, the mirror information was not populated.


jedsundwall (approved):
Thanks for jumping on this. I imagine we'll have to update this in the future when we let people choose other mirrors, but great to have this be functional again!

PR #156, opened 2025-09-21
feat: add account link popups

alukach:
## What I'm changing

Add a hover card to provide user with greater information about a given account when a user hovers over a link to an account.

https://github.com/user-attachments/assets/f97176fe-a8a8-4cc5-9dac-95086e8f1361

## How I did it

* Make use of Radix' [Hover Card component](https://www.radix-ui.com/primitives/docs/components/hover-card)
* Move all URLs to a urls util file to help better manage the growing number of internal links within the app (f74b97a2c53759581369a2295dac993d61b67330)
* Avoid problematic import of bright syntax highlighter by removing it from our components `index.ts` file (it can still be explicitely imported at its full path)

## How you can test it

Kick the tires over at https://source-cooperative-git-feat-add-account-popup-radiantearth.vercel.app

### DIFF
--- a/src/hooks/useKeyboardShortcuts.ts
+++ b/src/hooks/useKeyboardShortcuts.ts
@@ -1,5 +1,6 @@
 import { useEffect, useState } from 'react';
 import { useRouter, usePathname } from 'next/navigation';
+import { homeUrl, accountUrl, objectUrl } from "@/lib/urls";
 
 interface UseKeyboardShortcutsProps {
   onShowHelp?: () => void;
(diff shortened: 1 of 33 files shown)


src/hooks/useKeyboardShortcuts.ts
jedsundwall, line 1: Curious to know what's going on here. I can't tell! Are you improving our keyboard shortcuts?

jedsundwall (approved):
I love this very much. Makes me more anxious to get org avatars on here. Should I create an issue for that? I can see you've done some work on this in ProfileAvatar.tsx. I'd like us to display avatars whenever we can, including on https://github.com/source-cooperative/source.coop/blob/main/src/components/features/products/ProductListItem.tsx – they'd need to be smaller there, which I guess we have accommodations for given the size property in ProfileAvatarProps.

PR #159, opened 2025-09-22
pass theme parameter to parquet-table iframe

severo:
## What I'm changing

Pass a `theme` query parameter to the Parquet table iframe. See https://github.com/source-cooperative/parquet-table/pull/10

## How I did it

Get the theme from the context, and pass it to the iframe, using "light" as the default.

## How you can test it

It requires https://github.com/source-cooperative/parquet-table/pull/10 to be merged first. Then: go to a product with a parquet file, go the parquet page, see the embedded table: it should use the same theme as the page. If you change the theme in source.coop, it should change in the iframe too.

### DIFF
--- a/src/components/features/products/object-browser/ObjectPreview.tsx
+++ b/src/components/features/products/object-browser/ObjectPreview.tsx
@@ -1,17 +1,22 @@
 "use client";
 
 import { Box } from "@radix-ui/themes";
+import { useTheme } from 'next-themes';
 
 interface ObjectPreviewProps {
   sourceUrl: string;
 }
 
 const getIframeSrc = (sourceUrl: string) => {
+  const { resolvedTheme } = useTheme();
+  const theme = resolvedTheme === 'dark' ? 'dark' : 'light';
+
   switch (sourceUrl.split(".").pop()) {
     case "pmtiles":
+      // TODO(SL): add support for theme param in pmtiles.io?
       return `https://pmtiles.io/#url=${sourceUrl}&iframe=true`;
     case "parquet":
-      return `https://source-cooperative.github.io/parquet-table/?iframe&url=${sourceUrl}`;
+      return `https://source-cooperative.github.io/parquet-table/?url=${sourceUrl}&iframe=true&theme=${theme}`;
     default:
       return null;
   }


severo:
It adapts as expected:

Screenshot From 2025-09-22 12-20-16
Screenshot From 2025-09-22 12-19-34

src/components/features/products/object-browser/ObjectPreview.tsx
```diff
@@ -1,24 +1,29 @@
...
 interface ObjectPreviewProps {
   sourceUrl: string;
 }
 
-const getIframeSrc = (sourceUrl: string) => {
+const getIframeSrc = ({sourceUrl, theme}: ObjectPreviewProps & {theme: 'dark' | 'light'}) => {
   switch (sourceUrl.split(".").pop()) {
     case "pmtiles":
+      // TODO(SL): add support for theme param in pmtiles.io?
       return `https://pmtiles.io/#url=${sourceUrl}&iframe=true`;
```
severo, line 14: Do we want to pass it to pmtiles.io too?

cc @bdon

jedsundwall:
I'd like to test this but the PR on the parquet viewer repo is in draft mode.

alukach:
> I'm wondering if we should change the themes to 'source-dark' / 'source-light', or combine the tgeme with another way for the viewer to know it's beibg embedded in an iframe in source.coop so that it can adapt the theme accordingly (mainly for pmtiles.io)

I assumed this would be custom to each viewer and be written into the `getIframeSrc()` logic:
https://github.com/source-cooperative/source.coop/blob/344a42530bae2156d251f0ace998d3a4d328ddf5/src/components/features/products/object-browser/ObjectPreview.tsx#L10-L20

I assume the various viewers will expect arguments unique to that viewer, so there isn't much hope in regards to us establishing a convention that we would expect for them to utilize.

severo:
Yes, maybe it's not very important.

My idea was that a viewer could have specific light and dark themes when it knows it's embedded in source.coop. Otherwise, the default viewer's light or dark themes might use different palettes and feel slightly off in the iframe.

But there is no reliable way to know the parent's URL (https://stackoverflow.com/a/69049876/7351594), so, if we want that feature, we have three options:
- only specify the theme (`light` / `dark`), provided the viewer has themes that are aligned with source.coop,
- specific theme names (`source-light`, `source-dark`) and the viewer should be able to provide specific themes for them.
- pass another parameter (`parent="source.coop"`) for example... But I don't think it's better.

PR #188, opened 2025-12-10
fix: rework to use 1.91:1 image ratio

alukach:
## What I'm changing

As per [LinkedIn's docs](https://www.linkedin.com/help/linkedin/answer/a521928), the OpenGraph image should be 1.91:1 ratio.  This PR alters the layout of the images to better conform to that.

### Before

image

### After

image

> [!NOTE]
> The avatar / description is slightly different between preview builds and production, don't get hung up on those differences.


jedsundwall (approved):
thank you!

PR #217, opened 2026-02-06
feat: use stac-map for stac-geoparquet files

### LINKED ISSUE #216: [Proposed Feature] viewer for STAC GeoParquet
**Description of Feature:**

If a parquet file conforms to the STAC GeoParquet spec[^1], we should consider rendering a viewer that takes advantage of the unique capabilities of STAC GeoParquet (eg CQL2 filtering, spatial visualization).

STAC GeoParquet has emerged as an effective way to catalog large amounts of data in a manner that is both space efficient and efficient to query without requiring server-side logic. This would allow data providers to provide an index to their data alongside the data to allow consumers to efficiently find data that is of interest.

Related:

- https://github.com/englacial/xopr/issues/64
- https://github.com/source-cooperative/source.coop/issues/209

**Implementation**

DevelopmentSeed's STAC Map[^2] ([app](https://developmentseed.org/stac-map/)) is capable of performing this visualization today. The challenge comes from determining when this viewer should be loaded. To the best of my knowledge, there is no naming convention to signify that a Parquet file conforms to the STAC GeoParquet specification. As such, to determine if a file should be viewed as a STAC GeoParquet, we should likely:

1. Ensure that the file is a Parquet file. This may be as simple as looking for a `.parquet` extension (as we do today[^3] for our Parquet viewer) or analyzing its `content-type` header
2. Perform a quick "sniff" operation to check for compliance as efficiently as possible. @gadomski mentions[^4] that we could perform a `select stac_version, geometry from 'file.parquet'` with DuckDB WASM, however we don't currently have DuckDB WASM loaded within our frontend and I worry that doing so reactively when we see a Parquet file would lead to an undesirable/awkward lag. On the otherhand, I worry that loading the DuckDB WASM runtime pro-actively would add bloat to the frontend. Not sure what the best path forward is at this moment.

[^1]: https://stac-utils.github.io/stac-geoparquet/latest/spec/stac-geoparquet-spec/
[^2]: https://github.com/developmentseed/stac-map
[^3]: https://github.com/source-cooperative/source.coop/blob/106c8348e162db988998ed0353388328fc35f464/src/components/features/products/object-browser/ObjectPreview.tsx#L19-L23
[^4]: https://github.com/englacial/xopr/issues/64#issuecomment-3856846554

alukach:
## What I'm changing

Render STAC GeoParquets via STAC Map

closes #216

## How I did it

I've moved the object preview component to be a server-side component. I did this based on the assumption that loaded DuckDB server-side would be faster than client-side with WASM (this has yet to be verified). Now, when a Parquet file is to be previewed, we boot up DuckDB on the server and perform a quick check to test if it conforms to the STAC GeoParquet spec. If so, the STAC Map iframe is used. Otherwise, the traditional HyParquet viewer is used.

### Demo

The following videos demonstrate the length of time required to determine which viewer to load. Performance isn't amazing, but not terrible either.

https://github.com/user-attachments/assets/2ad38324-8ef4-4cfc-b210-5caf50a6dfdf

https://github.com/user-attachments/assets/d24ff961-c67f-4af2-aec7-1db84d9bab55

## How you can test it

* ✅ STAC GeoParquet (uses STAC Map): https://source-cooperative-git-feature-stac-geoparquet-radiantearth.vercel.app/alukach/alukach-experimentation/naip.parquet
* ❌ non-STAC GeoParquet (uses HyParquet): https://source-cooperative-git-feature-stac-geoparquet-radiantearth.vercel.app/alukach/alukach-experimentation/overture/2024-02-15-alpha-0/theme=admins/type=administrativeBoundary/0.parquet

### DIFF
--- a/src/components/features/products/object-browser/ObjectPreview.tsx
+++ b/src/components/features/products/object-browser/ObjectPreview.tsx
@@ -3,7 +3,6 @@ import "server-only";
 import { Box, Skeleton } from "@radix-ui/themes";
 import type { CSSProperties } from "react";
 import { DuckDBConnection } from "@duckdb/node-api";
-import { cacheLife } from "next/cache";
 
 interface ObjectPreviewProps {
   sourceUrl: string;
@@ -94,11 +93,6 @@ const getIframeAttributes = async (
 export async function ObjectPreview({
   sourceUrl: cloudUri,
 }: ObjectPreviewProps) {
-  "use cache";
-  cacheLife({
-    stale: 60 * 60 * 24 * 7, // Cache for 1 week
-    revalidate: 60 * 60, // Background-refresh every hour
-  });
   const iframeProps = await getIframeAttributes(cloudUri);
   if (iframeProps) {
     const { src, style } = iframeProps;
--- a/next.config.js
+++ b/next.config.js
@@ -23,9 +23,6 @@ const nextConfig = {
     },
   },
   serverExternalPackages: ["@duckdb/node-api"],
-  experimental: {
-    useCache: true,
-  },
 };
 
 export default nextConfig;
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
         "lucide-react": "^0.483.0",
         "md5": "^2.3.0",
         "mystjs": "^0.0.15",
-        "next": "16.1.6",
+        "next": "15.5.9",
         "next-themes": "^0.4.4",
         "nextjs-toploader": "^3.8.16",
         "node-fetch": "^3.3.2",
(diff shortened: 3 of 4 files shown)


src/components/features/products/object-browser/ObjectPreview.tsx
```diff
@@ -1,22 +1,54 @@
...
+    // Vercel: /tmp is the only writable location
+    await db.run("SET home_directory='/tmp'");
+    await db.run("SET extension_directory='/tmp/duckdb_extensions'");
+
+    const reader = await db.runAndReadAll(
+      `
+        SELECT stac_version
+        FROM read_parquet(?)
+        WHERE geometry IS NOT NULL
+        LIMIT 1
```
alukach: Unsure about this query. I intended for it to be something where the query would fail if it didn't have a `stac_version` or `geometry` column but I'm wondering if we should also ensure that the `stac_version` is greater than some semver value.

e.g. for `1.2.4` we could do:

```sql
WITH first_row AS (
    SELECT
        split(stac_version, '.') AS v
    FROM read_parquet(
        'file.parquet',
        columns = ['stac_version']
    )
    LIMIT 1
)
SELECT
    (
        CAST(v[1] AS INTEGER) > 1
        OR (
            CAST(v[1] AS INTEGER) = 1
            AND CAST(v[2] AS INTEGER) > 2
        )
        OR (
            CAST(v[1] AS INTEGER) = 1
            AND CAST(v[2] AS INTEGER) = 2
            AND CAST(v[3] AS INTEGER) >= 4
        )
    ) AS is_stac_version_ge_1_2_4
FROM first_row;
```

PR #243, opened 2026-03-24
fix: adjust item layout for better spacing and alignment

alukach:
## What I'm changing

Make featured product cards expand to height of row.

> [!NOTE]
> Not sure if we actually want this, but it's a UI idea that I wanted to try out

## How I did it

Honestly, this stuff (flex layout) is a bit beyond me, I let Claude take the wheel.

### Before

image

### After

image

### DIFF
--- a/src/components/features/products/ProductList.module.css
+++ b/src/components/features/products/ProductList.module.css
@@ -58,6 +58,11 @@
 
 .gridList > li {
   display: flex;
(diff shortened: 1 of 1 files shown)


jedsundwall (approved):
Yes, this is the way. We can go through some of these and fill out the descriptions to make them look better if there's a lot of white space.

PR #256, opened 2026-03-30
fix: create search_text field for performant search

alukach:
## What I'm changing

Add a `search_text` field to the product model to allow us to perform case insensitive searches on products.

## How I did it

1. update search to perform the `contains()` operation on the `search_text` field
2. update `ProductsTable` to populate `search_text` field on creates/upserts
3. create script to backfill existing products

> [!NOTE]
> The `contains()` operation works as an index scan on the public products index. Anything more performant will require moving away from DynamoDB.

## How you can test it

The backfill script has been run on staging. Search can be tested on this PR's preview branch: https://source-cooperative-git-fix-search-case-insensitive-radiantearth.vercel.app/products?search=ftw

### DIFF
--- a/src/lib/clients/database/products.ts
+++ b/src/lib/clients/database/products.ts
@@ -114,7 +114,6 @@ export class ProductsTable extends BaseTable {
     const expressionValues: Record<string, any> = {
       ":visibility": "public",
     };
-    const expressionNames: Record<string, string> = {};
     const filterParts: string[] = [];
 
     let keyCondition = "visibility = :visibility";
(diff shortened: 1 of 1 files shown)


scripts/backfill-search-text.ts
```diff
@@ -0,0 +1,164 @@
...
+  return [
+    product.title,
+    product.description,
+    product.account_id,
+    product.product_id,
+  ]
+    .filter(Boolean)
+    .join(" ")
+    .toLowerCase();
+}
```
tylere: This code is duplicated in src/lib/clients/database/products.ts#L21-L36
Suggest adding comments to warn they must be updated together (or abstract the logic into a utility function).

PR #276, opened 2026-04-01
feat: add city country to globe labels

alukach:
## What I'm changing

For better or worse, I keep futzing with this globe.

Given that it's difficult to see land masses on the globe, I think it's a good idea to display the long-form name of each Cloudflare colocation center.

image

## How I did it

1. Get locations from https://github.com/Netrvin/cloudflare-colo-list
2. Augment locations with lat/lon with AI generated values
3. Rm unused variables

### DIFF
--- a/src/components/features/globe/LiveGlobe.module.css
+++ b/src/components/features/globe/LiveGlobe.module.css
@@ -68,6 +68,12 @@
   overflow: hidden;
   text-overflow: ellipsis;
 }
+.popupLocation {
+  margin-bottom: 4px;
+  color: var(--gray-11, #6b6b6b);
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
 .popupLink {
   color: inherit;
   text-decoration: none;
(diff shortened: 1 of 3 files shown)


jedsundwall (approved):
Very elegant. I like it a lot. I wonder if anyone will complain about privacy stuff. Like what happens when we find out that someone in a small town somewhere is looking at a product? 

I'm approving, but let me know if we're able to set this to only name the closest large (by some definition) city.

alukach:
> I wonder if anyone will complain about privacy stuff. Like what happens when we find out that someone in a small town somewhere is looking at a product?

I initially had the same concern but it's important to note that these locations are NOT the location of the requesting user (we do have access to their IP or their country, but don't currently retain that). Instead, these are the locations of servers within [the Cloudflare network](https://www.cloudflare.com/network/) that process the requests, so they represent a crude approximation of the requester's location (ie generally the closest large city, as you suggested).

PR #294, opened 2026-04-20
fix: set canvas background to match page background

alukach:


alukach:
I'm honestly not sure if this makes any difference.

PR #300, opened 2026-04-30
fix: make username lookups case-insensitive

### LINKED ISSUE #299: [Bug] Username lookups should be case insensitive
**Description of Bug:**

When I type in a username in the email address, the lookup is case sensitive. We should do case insensitive lookups instead.

**Steps to Reproduce:**

Navigate to: https://source.coop/ALukach

**Expected Behavior:**

Serves user https://source.coop/alukach

**Actual Behavior:**

Account not found

**Screenshots:**

**Additional Context:**

alukach:
Normalize account_id and product_id to lowercase before DynamoDB queries so that URLs like /ALukach resolve correctly to the /alukach account. Also redirects mixed-case account page URLs to their canonical lowercase form.

Closes #299

### DIFF
--- a/src/app/(app)/[account_id]/page.tsx
+++ b/src/app/(app)/[account_id]/page.tsx
@@ -11,7 +11,7 @@
  */
 
 import { Metadata } from "next";
-import { notFound } from "next/navigation";
+import { notFound, redirect } from "next/navigation";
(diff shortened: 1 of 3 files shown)


alukach:
@claude review this PR. Do we need to normalize all of the already stored account_ids?

PR #301, opened 2026-05-01
fix: validate bbox struct type before routing to STAC map viewer

### LINKED ISSUE #297: [Bug] Binder Error reported by Parquet viewer
**Description of Bug:** The interactive map viewer shows an error when viewing a parquet dataset.

**Steps to Reproduce:**
- Step 1  View https://source.coop/maxar/maxar-opendata/maxar-opendata.parquet

**Expected Behavior:**

Interactive map showing the parquet dataset.

**Actual Behavior:**

Map displays an error in an overlay:

`Error: Binder Error: Cannot extract field 'xmin' from expression "bbox" because it is not a struct, union, map, or json`

**Screenshots:**

Image

**Additional Context:**

tylere:
Files like maxar-opendata.parquet store the bbox column as a list/array rather than a struct with xmin/ymin/xmax/ymax fields. The STAC map viewer assumes bbox is a struct and throws a DuckDB Binder Error when it tries to access bbox.xmin on a non-struct column.

Extend the isStacGeoParquet schema check to also require that if a bbox column is present, it must have an xmin sub-field (indicating a proper struct). Files without a struct bbox fall back to the parquet-table viewer.

Closes #297

### DIFF
--- a/src/components/features/products/object-browser/ObjectPreviewExternal.tsx
+++ b/src/components/features/products/object-browser/ObjectPreviewExternal.tsx
@@ -16,11 +16,21 @@ const isStacGeoParquet = async (sourceUrl: string): Promise<boolean> => {
     await db.run("SET home_directory='/tmp'");
     await db.run("SET extension_directory='/tmp/duckdb_extensions'");
 
+    // Check for required STAC columns and validate bbox is a struct (has xmin sub-field).
+    // Some files (e.g. Maxar) store bbox as a list/array rather than a struct, which causes
+    // the STAC map viewer to throw "Cannot extract field 'xmin' from expression 'bbox'".
     const reader = await db.runAndReadAll(
-      `SELECT count(*) > 0 FROM parquet_schema(?) WHERE name IN ('stac_version', 'geometry') GROUP BY file_name HAVING count(*) = 2`,
+      `WITH schema AS (SELECT * FROM parquet_schema(?))
+       SELECT
+         count(*) FILTER (WHERE name IN ('stac_version', 'geometry')) = 2
+           AND (count(*) FILTER (WHERE name = 'bbox') = 0
+                OR count(*) FILTER (WHERE name = 'xmin') > 0)
+         AS is_valid
+       FROM schema`,
       [sourceUrl],
     );
-    return reader.getRows().length > 0;
+    const rows = reader.getRows();
+    return rows.length > 0 && (rows[0][0] as boolean);
   } catch (error) {
     LOGGER.error(`Error checking for STAC GeoParquet: ${error}`, {
       operation: "isStacGeoParquet",


tylere:
Manually confirmed this fallback works for the Maxar dataset:
https://source-cooperative-git-claude-issue-297-202-80957d-radiantearth.vercel.app/maxar/maxar-opendata/maxar-opendata.parquet

alukach (commented):
There is a higher level question regarding if it's better to have our app let STAC handle/possibly-error-on broken STAC Geoparquet files or if we should silently fallback to the genetic parquet viewer as this PR does. 

By virtue of the `stac_version` column's presence, the maxar dataset is likely attempting to implement stac-geoparquet however it is encoded incorrectly (or, at least, in a way that the current version of stac map doesn't support). I could see confusion on the part of the data provider when they see the standard parquet viewer while expecting to see stac map. An astute observer may be able to identify that it is the incorrect `bbox` field causing the error, but I don't think that's very obvious. Perhaps we should still push to stac map with the mindset that the parquet files has met our requirements (existence of `geometry` and `stac_version`) and focus our energy on improving error handling within the viewer.

I'd be curious to hear @gadomski's thoughts on this as well.

tylere:
@alukach that is a good point. If stac-map provided clear guidance on what went wrong and how to address it (e.g. https://github.com/developmentseed/stac-map/issues/427), then this PR would hide that useful info. In that case, this PR should be reverted (if previously merged).

gadomski:
Yeah, in the interest of permissive reading, I'd like to support this file in **stac-map** if possible. @alukach our current `iframe` strategy makes it hard/impossible for **stac-map** to signal that it's errored back to source.coop ... however, my recent push towards providing `StacMap` as a React component (https://github.com/developmentseed/stac-map/pull/416) would make it possible, at least in a React context, to have an error handler.

I'll take a look at https://github.com/developmentseed/stac-map/issues/427 today or Monday and see if I can make a fix.

gadomski:
This should be fixed as of https://github.com/developmentseed/stac-map/releases/tag/v2.0.0-alpha.3, via this PR: https://github.com/developmentseed/stac-map/pull/431.

@alukach I've thought a bit about how easy it would be for me to rug-pull source.coop behavior by updating **stac-map** in an unexpected way. Would you want me to publish "permalinks" to **stac-map** versions (e.g. https://developmentseed.org/stac-map/v1.6.1) so you can choose when to update?

alukach:
> Would you want me to publish "permalinks" to stac-map versions (e.g. https://developmentseed.org/stac-map/v1.6.1) so you can choose when to update?

@gadomski yeah, we've talked about this before ([link](https://github.com/source-cooperative/source.coop/pull/217#discussion_r2775167335)). To repeat:

> If you're up to add versioning to stac-map, we'd take advantage if needed, but it's not a high priority at this point.

gadomski:
🙈 thanks for reminding me, totally forgot about that conversation

alukach:
@tylere feel free to push us in a direction regarding @gadomski's thoughts.

Regarding the core issue itself, it seems like Pete's fix did the trick 🎉 

https://github.com/user-attachments/assets/2a47251b-01e9-4fe3-b4e3-57f6acbab5b1

Going to close this and #297

tylere:
I think @alukach's comment from February still holds true...

> If you're up to add versioning to stac-map, we'd take advantage if needed, but it's not a high priority at this point.

@gadomski thanks for filing https://github.com/developmentseed/stac-map/issues/433, we'll watch it.

PR #305, opened 2026-05-07
fix: Escape special characters in schema.org JSON-LD metadata

tylere:
## What I'm changing

This PR escapes special HTML characters (`<`, `>`, `&`) in schema.org JSON-LD metadata to prevent potential XSS vulnerabilities and ensure proper rendering of structured data on product and account pages.

## How I did it

Updated both `ProductSchemaMetadata.tsx` and `AccountSchemaMetadata.tsx` to escape special characters in the JSON-LD script content:
- `<` is replaced with `\u003c`
- `>` is replaced with `\u003e`
- `&` is replaced with `\u0026`

This follows security best practices for embedding JSON-LD in HTML documents and prevents any user-supplied data within the schema from breaking out of the script tag.

Also bumped the version to 1.3.0 in `package.json` to reflect the release.

## How you can test it

1. Navigate to a product page and inspect the page source to verify the schema.org metadata in the `<script type="application/ld+json">` tag contains escaped characters where applicable
2. Navigate to an account/profile page and verify the same escaping is applied
3. Verify that structured data is still correctly parsed by checking with Google's Rich Results Test or similar schema validation tools
4. Confirm no visual or functional changes to the pages themselves

https://claude.ai/code/session_01UCrzLXdCiDYw7WKmZ6Ueot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [1.3.0](https://github.com/source-cooperative/source.coop/compare/v1.2.0...v1.3.0) (2026-05-07)
+
+
+### Features
+
+* add ROR ID field to organization profile edit form ([#280](https://github.com/source-cooperative/source.coop/issues/280)) ([00f3ff8](https://github.com/source-cooperative/source.coop/commit/00f3ff8c22a91bc0347e75e20a77f8edc6645964))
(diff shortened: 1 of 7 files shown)


alukach (approved):
I admittedly don't know how much of an issue this is being that we have `JSON.stringify()` but I don't think this would hurt.

PR #344, opened 2026-06-07
fix: re-render auth UI after form-submission redirects

alukach:
## Problem

After submitting certain forms (e.g. creating a product), the user is redirected to a new page where they appear **logged out**. Reloading the browser shows them correctly logged in again.

## Root cause

This is the Next.js App Router shared-layout / Router Cache gotcha — the session is never actually lost, only the UI is stale.

- The auth UI is rendered server-side in the shared `src/app/(app)/layout.tsx` → `Navigation` → `AuthButtons`, which reads the session via `getPageSession()` / Ory `getServerSession()`.
- Redirecting server actions finish with `redirect()`, which performs a **soft** (client-side) navigation.
- On a soft navigation, Next.js reuses the **client Router Cache** entry for shared layout segments and does **not** re-execute the layout's server component. So `AuthButtons` is never re-rendered and a stale (logged-out / skeleton) snapshot of the layout is shown.
- A full browser reload is a hard navigation that re-renders the layout server-side with the live Ory cookie — hence "reload fixes it".

This is why only the **redirecting** actions exhibit it (`createProduct`, `createAccount`); the update/membership actions call `revalidatePath(...)` and stay on the same page.

## Fix (Solution 1)

Call `revalidatePath("/", "layout")` immediately before each `redirect()` so the soft navigation invalidates and re-renders the root layout with the current session.

- `src/lib/actions/products.ts` — `createProduct`
- `src/lib/actions/account.ts` — `createAccount`

Both files already imported `revalidatePath`, so the change is minimal and matches the existing `revalidatePath` usage in the non-redirecting actions.

## Testing

- `npx jest src/lib/actions` → 21 passed
- Type check clean (only a pre-existing unrelated `baseUrl` deprecation warning)

https://claude.ai/code/session_01TeDE1T3qU4Ekw5MET3BdWf

---
_Generated by [Claude Code](https://claude.ai/code/session_01TeDE1T3qU4Ekw5MET3BdWf)_

### DIFF
--- a/src/components/core/DynamicForm.tsx
+++ b/src/components/core/DynamicForm.tsx
@@ -3,6 +3,7 @@
 import React, { useActionState } from "react";
 import { Button, Text, Flex } from "@radix-ui/themes";
 import Form from "next/form";
+import { useRouter } from "next/navigation";
 
 export interface FormField<T extends Record<string, any>> {
   label?: string;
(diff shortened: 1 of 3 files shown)


alukach:
Verified fixed 

## Before
https://github.com/user-attachments/assets/393c3d98-f658-498f-9f39-c2af1fef076a

## After

https://github.com/user-attachments/assets/157b5e59-444b-46b2-80b2-1e79d76fae46

PR #348, opened 2026-06-09
feat(auth): verify email status on every authenticated page load

alukach:
## What

Adds a `VerificationBanner` to the `(app)` layout so that on **every authenticated page load**, while the signed-in user's account is **not yet marked verified in DynamoDB**, we reconcile verification state with Ory and surface the right banner.

Decision flow (only runs while our DB record is unverified — once synced it short-circuits to nothing):

| State | Behavior |
|-------|----------|
| Unauthenticated / no individual account / already verified in our DB | Render nothing |
| Unverified in our DB **and** on Ory | Render a **"please verify your email"** reminder with a link to Ory verification |
| Unverified in our DB **but** verified on Ory | **Sync our DynamoDB record** to match Ory, then render a **"thank you for verifying"** banner. The next load reads the synced record and renders nothing. |

## How

- New server component `src/components/features/auth/VerificationBanner.tsx`, mounted in `src/app/(app)/layout.tsx` inside a `<Suspense fallback={null}>` so the session/DB read streams in without blocking navigation.
- New pure, unit-tested helpers in `src/lib/accounts/email-verification.ts`:
  - `isEmailVerifiedInDb(account)`
  - `isEmailVerifiedInOry(orySession)`
  - `oryAddressesToAccountEmails(orySession)` — projects Ory's `verifiable_addresses` onto our `AccountEmail` shape.
- Refactored the existing `email-verified` page to reuse these helpers, removing the duplicated mapping. (Minor correctness improvement: `verified_at` is now only set for addresses that are actually verified.)

The DB sync is written during render of the banner, matching the existing precedent in the `email-verified` page. It is idempotent and only fires in the narrow window where Ory is verified but our record is stale.

## Testing

- `src/lib/accounts/email-verification.test.ts` — 8 passing unit tests covering the DB/Ory predicates and the address mapping (primary flag, timestamp passthrough, unverified `verified_at` omission, empty input).
- `tsc --noEmit`: 0 errors.
- `next lint` on changed files: clean.

### DIFF
--- a/src/app/(app)/email-verified/page.tsx
+++ b/src/app/(app)/email-verified/page.tsx
@@ -9,6 +9,10 @@ import { redirect } from "next/navigation";
 import { accountUrl, homeUrl, onboardingUrl, verifyEmailUrl } from "@/lib/urls";
 import { getPageSession } from "@/lib/api/utils";
 import { accountsTable } from "@/lib";
+import {
+  isEmailVerifiedInOry,
+  oryAddressesToAccountEmails,
+} from "@/lib/accounts/email-verification";
 import Link from "next/link";
 
 export default async function EmailVerifiedPage() {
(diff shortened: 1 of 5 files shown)


alukach:
Closing in favor of #347, which now incorporates this PR's architecture — the Suspense-isolated `VerificationBanner` and the pure, unit-tested helpers in `src/lib/accounts/email-verification.ts` (including the `verified_at`-only-when-verified fix) — on top of #347's broader UX cleanup. Superseded by #347.

PR #350, opened 2026-06-09
feat(admin): admin email-to-profile lookup view

alukach:
## Summary

Adds an **admin area** to collect admin tools, with the first tool being a **user lookup**: an admin submits an email, it's resolved against Ory, and (if a matching user with a source.coop profile exists) the admin is redirected to that user's profile. Otherwise they're told the email wasn't found.

## Structure (extensible)

- **`ADMIN_TOOLS` registry** (`src/components/features/admin/tools.ts`) — single source of truth for admin tools.
- **`/admin` layout** gates the whole segment by `isAdmin`, so child pages assume an admin visitor.
- **`/admin` home page** lists tools as linked cards.
- **Account dropdown → Admin submenu** (nested `DropdownMenu.Sub`, admin-only) is driven by the registry and links to the admin home.

Adding a tool later = one registry entry + its route under `src/app/(app)/admin/`.

## User Lookup tool

- `/admin/user-lookup` page (width-constrained form).
- `lookupUserByEmail` server action: validates the email (zod), re-checks `isAdmin` (actions aren't covered by the layout gate), calls `getOryIdentityIdByEmail` (`GET /admin/identities?credentials_identifier=<email>`), maps the Ory identity to a source.coop account via `accountsTable.fetchByOryId`, and redirects to `/<account_id>`.
- Messaging distinguishes **not in Ory** vs **in Ory but no source.coop profile**.

## Notes

- Ory admin calls read credentials via `CONFIG` (`ORY_PROJECT_API_KEY`). The pre-existing `getEmail` helper was reading an unset `ORY_ACCESS_TOKEN`, sending `Bearer ` and 401ing on every admin call; this PR fixes that helper too.

## Testing

- `npx tsc --noEmit` — clean
- `admin.test.ts` — 6/6 pass; `DropdownSection.test.tsx` (unit) passes
- Verified against live Ory: a known email resolves to its identity (HTTP 200, 1 match).
- Pre-existing `DropdownSection.integration.test.tsx` failure (a `React.Children.only` / duplicated `@radix-ui/react-slot` issue) is unrelated and present on `main`.


tylere:
Comments from Cursor/Fable 5 review:

# Overall

Clean, small, and well-tested for its size. The action correctly re-checks isAdmin (with the right reasoning — layouts don't cover Server Actions), nothing sensitive is logged (the audit log records account_id and looked_up_by, not the email), and I verified FormState.redirectTo is real on current main (DynamicForm does router.refresh() + router.push() on success — my local checkout was stale, so don't be alarmed if you don't see it locally). Two things worth addressing:

# Findings

## **1. The layout comment encodes a dangerous assumption about Next.js (worth fixing now, before more tools exist).**

layout.tsx
Lines 15-18
// Gates every route under /admin and presents the admin tools in a settings-
// style layout (left sidebar menu + content area), matching the account and
// product edit views. Child pages can assume the visitor is an admin and focus
// on their own content.
In the App Router, layouts and pages render in parallel — a layout returning <NotAuthorizedPage /> prevents the children's output from being sent, but it does not stop a child page's server code (data fetches, generateMetadata) from executing. Today that's harmless: user-lookup/page.tsx is static UI and the action self-checks. But "child pages can assume the visitor is an admin" invites a future tool to fetch sensitive data in its page component with no check — and that fetch will run for non-admins. Next.js' own guidance is to authorize close to the data, not in layouts. Suggestion: a tiny requireAdmin() helper called in each admin page (and keep the layout check for UX), and reword the comment.

## **2. Ory outage is reported as "no user found" (medium).**

getOryIdentityIdByEmail returns null for any non-OK response — expired ORY_PROJECT_API_KEY, Ory downtime, a 429 — and the action then tells the admin "No user found in Ory for [x@y.com](mailto:x@y.com)". That's a wrong answer, not a degraded one, and admins will act on it (e.g. concluding an account doesn't exist). The status is logged at warn, but the UI should distinguish "lookup failed, try again" from "not found" — e.g. throw on non-OK (or return a discriminated result) and surface a generic failure message. A test for the non-OK path would lock that in.

## Minor

- PR description drift: the body says the /admin home "lists tools as linked cards," but admin/page.tsx just redirects to the first tool. Consequently AdminTool.description is never rendered, and adminUrl() appears unused. Either restore the card view or trim the dead field/helper and update the description.
- Multiple Ory matches: identities[0] is taken silently. credentials_identifier should be effectively unique for emails, but logging when length > 1 would surface surprises cheaply.
- NotAuthorizedPage (200) vs notFound(): the admin area's existence is revealed to non-admins. Probably fine for this app — just a conscious choice to confirm.
- Merge coordination with PR #283: both PRs touch src/lib/api/utils.ts, and both apply the same getEmail fix (CONFIG.auth.api.backendUrl / CONFIG.auth.accessToken). #283 also restructures that file heavily, so whichever lands second gets a conflict — trivial, but worth sequencing deliberately.
- initialState: any / FormState<any> is loose, but consistent with the existing form actions, so fine to leave.

## What's good

- The registry pattern (ADMIN_TOOLS) keeps the dropdown, sidebar, and routes in sync from one place — adding a tool really is one entry + one route.
- Test coverage hits the right paths: non-admin rejection, zod validation, Ory-miss vs no-profile distinction, success redirect, and email trimming.
- The drive-by getEmail fix (previously sending Bearer with an unset ORY_ACCESS_TOKEN) is small, correct, and explained in the PR body.

alukach:
I think all of these findings are non-critical, particularly for an admin-only view. Went ahead and merged to reduce PR count (pushing focus to more mission-critical changes)

PR #364, opened 2026-06-18
feat(admin): data connection management (CRUD, S3/Azure/GCP/R2, product mirrors)

alukach:
## What I'm changing

Adds a full **Data Connection admin** (browse / create / edit / delete) plus product-side mirror management. This reworks the closed #342 onto the admin menu from #350 and the data-connection model from #332 (now merged to `main`).

Admins can now:
- Manage data connections from **Admin → Data Connections**.
- Configure **S3, Azure, GCP (GCS),** and **S3-compatible backends (Cloudflare R2, MinIO, Ceph)**.
- See which products use a given connection, and jump from a product's mirrors straight to that connection's admin page.
- Attach/detach connections and set the primary mirror from a product's **Data Connections** tab.

Supersedes #342 · builds on #350 (admin menu) and #332 (model) · follow-up: #365 (enforce `required_flag`).

## How I did it

**Admin-menu integration (#350).** Registered "Data Connections" in the `ADMIN_TOOLS` registry, which drives both the `/admin` sidebar and the account-dropdown Admin submenu — no standalone admin layout. New routes live under `src/app/(app)/admin/data-connections/` (list / create / `[id]` edit).

**The form.** `DataConnectionForm` is a client form with provider- and auth-type-conditional fields (the shared `DynamicForm` can't express the discriminated `details`/`authentication` shapes). Every field carries an inline description, and the Authentication Type description changes with the selected mechanism:
- **S3** → ECS task role / access key / web-identity role / local, plus an optional **Endpoint** field + `auto` region for R2/MinIO/Ceph.
- **Azure** → SAS token / workload identity.
- **GCP** → workload identity (provider + service account).

**Server actions** (`data-connections.ts`, `product-mirrors.ts`). Each re-checks authorization (actions aren't covered by the layout gate), validates against the Zod schema, and maps Zod issues to **leaf** field names so nested `details.*` / `authentication.*` errors render next to their inputs. Navigation uses the `FormState.redirectTo` client-nav pattern (consistent with the rest of the app; avoids the stale Router Cache from a server `redirect()`).

**Security.** Storage secrets (access keys, SAS tokens) are never serialized to the client: a `redact` module strips them before any `DataConnection` is handed to a client component, and the edit form omits secret inputs (blank = keep existing on update).

**Data layer.** Added `DataConnectionsTable.delete`; fixed `update` (alias the reserved word `name`; `REMOVE` undefined optionals instead of referencing values the document client strips); `create` uses a conditional put so a concurrent create can't silently overwrite a live connection. Added `ProductsTable.listProductsByConnectionId` — a full scan + app-side filter (admin-scale; a denormalized index + GSI is the noted upgrade path) backing the "Products using this connection" panel.

**Mirrors.** `ProductMirrorsManager` (product edit tab) lets admins add/remove/set-primary; `addProductMirror` maps provider → `storage_type` (`gcp → gcs`), and the product file page now guards against an empty/stale `primary_mirror`.

The branch was hardened via a 5-agent review that caught secret leakage to the client, a broken DynamoDB `update`, a dropped SAS-token field, and a mirror-removal crash.

**Data Proxy impact.** No breaking changes to existing connection shapes. The S3 `endpoint` field is additive. **GCP Workload Identity federation is configurable here but not yet implemented in the proxy/multistore** (schema scaffolding, same status as Azure), so GCP connections won't be functional downstream until that lands; R2 via static keys is functional.

## How you can test it

1. As an **admin**, open **Admin → Data Connections** (or the account dropdown → Admin).
2. **Create** — try S3 (with and without an Endpoint for R2, region `auto`), Azure, and GCP. Switch provider/auth type and confirm the fields + descriptions update. Submit an invalid value (e.g. a malformed Role ARN) and confirm the error renders next to that field.
3. **Edit** — reopen a connection that has secrets: secret inputs are blank with a "leave blank to keep" hint, and saving without re-entering them preserves the stored values.
4. **Delete** — confirm the dialog and the redirect back to the list.
5. **Product → Data Connections tab** — add a connection, set it primary, remove it; as an admin, use the ↗ link next to a connection ID to open it in the admin.
6. **Connection edit page** — confirm the **Products using this connection** list (with a Primary badge).

Automated checks (all green): `npm run type-check`, `next lint`, and `npx jest src/lib/actions/data-connections.test.ts src/lib/actions/product-mirrors.test.ts` (27 tests).

### DIFF
--- a/src/app/(app)/[account_id]/[product_id]/(product)/[[...path]]/page.tsx
+++ b/src/app/(app)/[account_id]/[product_id]/(product)/[[...path]]/page.tsx
@@ -84,12 +84,14 @@ export default async function ProductPathPage({ params }: PageProps) {
         | { primaryMirror: ProductMirror; dataConnection: DataConnection }
         | undefined;
 
+      // primary_mirror may be empty or stale (e.g. its mirror/connection was
+      // removed), so guard before dereferencing.
       const primaryMirror =
         product.metadata.mirrors[product.metadata.primary_mirror];
-      const dataConnection = await dataConnectionsTable.fetchById(
-        primaryMirror.connection_id,
-      );
-      if (dataConnection) {
+      const dataConnection = primaryMirror
+        ? await dataConnectionsTable.fetchById(primaryMirror.connection_id)
+        : null;
+      if (primaryMirror && dataConnection) {
         connectionDetails = { primaryMirror, dataConnection };
       }
 
--- a/src/app/(app)/admin/data-connections/[data_connection_id]/page.tsx
+++ b/src/app/(app)/admin/data-connections/[data_connection_id]/page.tsx
@@ -1,11 +1,13 @@
 import { Metadata } from "next";
-import { Flex, Heading } from "@radix-ui/themes";
+import { Suspense } from "react";
+import { Flex, Heading, Text } from "@radix-ui/themes";
 import { notFound } from "next/navigation";
 import { dataConnectionsTable } from "@/lib/clients";
 import {
   DataConnectionForm,
   DeleteDataConnectionButton,
 } from "@/components/features/data-connections";
+import { ConnectionUsage } from "@/components/features/data-connections/ConnectionUsage";
 import { toEditableDataConnection } from "@/components/features/data-connections/redact";
 
 export const metadata: Metadata = {
@@ -40,6 +42,16 @@ export default async function EditDataConnectionPage({
         mode="edit"
         dataConnection={toEditableDataConnection(dataConnection)}
       />
+
+      <Suspense
+        fallback={
+          <Text size="2" color="gray">
+            Loading product usage…
+          </Text>
+        }
+      >
+        <ConnectionUsage connectionId={dataConnection.data_connection_id} />
+      </Suspense>
     </Flex>
   );
 }
--- a/src/app/(app)/admin/data-connections/[data_connection_id]/page.tsx
+++ b/src/app/(app)/admin/data-connections/[data_connection_id]/page.tsx
@@ -1,3 +1,4 @@
+import { Metadata } from "next";
 import { Flex, Heading } from "@radix-ui/themes";
 import { notFound } from "next/navigation";
 import { dataConnectionsTable } from "@/lib/clients";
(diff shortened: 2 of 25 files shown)


alukach:
Addressed the review in `9ce14ca`:

- **1 (empty credentials)** — moved the constraint into the schema: `access_key_id` / `secret_access_key` / `sas_token` are now `z.string().min(1, …)`. Removed the redundant `missingSecretErrors` guard, so validation is self-contained and covers any future caller. The edit "leave blank to keep" flow is unaffected (the stored secret is filled in before validation).
- **2 (misleading sort)** — dropped the descending-by-ID sort (and the defensive copy that only existed to sort safely) from `listAll`; all three callers re-order as they need.
- **4 (`replace` → `replaceAll`)** — done; repeated `{{repository.*}}` placeholders now fully substitute.
- **5 (GCS trailing dash)** — picker now omits the ` - region` segment when region is empty. `redact.ts` already returns `""` for GCS, so only the display string changed.
- **3 (mirror optimistic lock)** — deliberately deferred. The race is admin-only, low-frequency, and "unlikely" as noted; closing it means threading a `ConditionExpression` through the shared `productsTable.update()` (currently a fixed `UpdateExpression`), which isn't worth it for the blast radius (one re-doable mirror toggle). Left a comment at the top of `product-mirrors.ts` naming the ceiling and the compare-and-swap upgrade path. Happy to implement if you'd prefer it closed now.

`npm run type-check`, `next lint`, and the 31 tests across `data-connections` / `product-mirrors` / `route` suites all pass.

alukach:
Resolved the last open finding (#3, mirror optimistic lock) in `c8664ef`:

- `productsTable.update()` now takes an optional `expectedUpdatedAt` and, when set, writes under `ConditionExpression: "updated_at = :expected_updated_at"` — a compare-and-swap. Existing single-arg callers keep last-write-wins, so nothing else changes.
- `addProductMirror`, `removeProductMirror`, and `setPrimaryMirror` pass the `updated_at` they read. If a concurrent admin edit bumped it, the write fails with `ConditionalCheckFailedException` and the action returns "This product was modified by someone else. Please reload and try again." instead of silently clobbering the other change.
- Added a test covering both the lock arg and the conflict path.

All 5 review findings are now addressed. `npm run type-check`, `next lint` (no new warnings), and the data-connections / product-mirrors / products suites pass (34 tests).

Issue #328, opened 2026-06-02
feat(admin-ui): create/edit federated data connection (subject_scope)


alukach:
Part of the **Federated backend authentication** epic (app-side).

### What
Admin/owner UI to create and edit a federated (`S3WebIdentityRole`) data connection.

### Fields
- `role_arn` (required) — `arn:aws:iam::<acct>:role/<role>`
- `audience` (default `source-coop-data-proxy`, advanced)
- `subject_scope` — radio: **Connection** (default) / Account / Product, with inline help explaining the caching vs. isolation trade-off
- `session_duration_secs` (advanced, 900–43200)

### Notes
- No secret input for this type (contrast with `S3AccessKey`).
- Show the rendered `sub` pattern preview (e.g. `scv1:conn:{id}:account:{acct}*`) so owners can author their IAM trust policy.

### Acceptance
- [ ] Create + edit federated connection end-to-end
- [ ] `subject_scope` selector with help text
- [ ] `sub` pattern preview shown for trust-policy authoring

alukach:
**Blocked by:** #326 (schema). Reads config back via #327.

alukach:
## Design update — 2026-06-06: `subject_scope` selector → `cache_level` selector

This supersedes the `subject_scope` design across the federated-backend-auth epic.

**1. One fixed subject shape.** The OIDC `sub` is now *always* product-grained: `scv1:conn:{connection_id}:{account_id}/{product_id}`. `render_subject` is branchless; the three-shape contract is dropped.

**2. Granularity moves to a per-connection `cache_level` enum (`product | account | system`)** controlling *only* the cache key — never the subject: `product` → `(conn, account, product)` · `account` → `(conn, account)` · `system` → `(conn)`.

**3. `cache_level` is a safety assertion, not a perf knob.** Coarser-than-`product` caching reuses one product's credential for its siblings without re-checking their subject at STS, so it must be no coarser than what the trust policy enforces. Default `product` (always safe).

---

**Changes for this ticket (#328 — admin UI):**

- The selector is now **`cache_level`** (radio: **Product** (default) / **Account** / **System**), not `subject_scope`.
[...]

Issue #325, opened 2026-06-02
Epic: Federated backend authentication (app-side — schema, API, UI, onboarding)


alukach:
## Epic: Federated backend authentication — app-side

Enable the data proxy to communicate with **private** storage backends via **OIDC federation**, so the proxy assumes a customer-owned role at request time instead of Source storing long-lived bucket secrets. This epic covers the `source.coop` (app/API/UI) half; the proxy-side half lives in [`source-cooperative/data.source.coop`](https://github.com/source-cooperative/data.source.coop) (see its companion epic).

### Why

Today the proxy resolves a product's backend bucket/region/prefix but hardcodes `skip_signature: true` / `anonymous_access: true` (`data.source.coop/src/registry.rs`), so only anonymous-public buckets work. `DataConnection.authentication` secrets are stored but never consumed. Federation lets the proxy authenticate to private backends while **storing only a role ARN** — no secret at rest in DynamoDB or at the edge.

### Key idea

The proxy is already an OIDC IdP (serves `/.well-known/openid-configuration` + JWKS). A `DataConnection` stores an AWS role ARN + audience + `subject_scope`. At resolve time the proxy mints a short-lived RS256 assertion and exchanges it via `AssumeRoleWithWebIdentity` for temporary, prefix-scoped credentials.

### Caching constraint that shapes the schema

Anything varied at AssumeRole time is baked into the credential, so granularity ⟺ cache cardinality. `subject_scope` (`connection` | `account` | `product`) is therefore a **per-connection** setting; the shared Source-owned connection (~99% of products) uses `connection` scope and keeps exactly one cached credential — zero regression.

### Scope (sub-issues)

App-side schema, secret-less API exposure, admin UI, proxy service-identity recognition, and customer onboarding IaC. S3 first; Azure/GCS are fast-follow.

### Reference

Design doc: `source.coop/docs/plans/2026-06-02-federated-backend-auth-design.md` (local working doc; `docs/plans/` is gitignored).

alukach:
Companion proxy-side epic: source-cooperative/data.source.coop#137

alukach:
## ✅ Federation approach validated end-to-end

[...]

PR #375, opened 2026-06-19
feat(data-connection): derive federated cache_level from connection owner (#326)

### LINKED ISSUE #326: feat(data-connection): add S3WebIdentityRole auth type to schema
Part of the **Federated backend authentication** epic (app-side).

### What
Add an `S3WebIdentityRole` authentication type to the `DataConnection` discriminated union. Unlike `S3AccessKey`, it carries **no secret** — only a role ARN and federation parameters.

### Where
`src/types/data-connection.ts`

```ts
export enum DataConnectionAuthenticationType {
  // ...existing
  S3WebIdentityRole = "s3_web_identity_role",
}

export const S3WebIdentityRoleAuthenticationSchema = z.object({
  type: z.literal(DataConnectionAuthenticationType.S3WebIdentityRole),
  role_arn: z.string(),                                  // arn:aws:iam::<acct>:role/<role>
  audience: z.string().default("source-coop-data-proxy"),
  subject_scope: z.enum(["connection", "account", "product"]).default("connection"),
  session_duration_secs: z.number().min(900).max(43200).default(3600),
});
```
Add to `DataConnectionAuthenticationSchema` union + OpenAPI registration.

### Acceptance
- [ ] New variant parses/validates; existing variants unaffected
- [ ] `subject_scope` defaults to `connection`
- [ ] Type-check + unit tests pass

alukach:
Closes #326. Completes the **schema** half of the federated-backend-auth epic (#325).

## What changed (vs the first revision of this PR)

The [2026-06-06 design pivot](https://github.com/source-cooperative/source.coop/issues/325#issuecomment-4640367748) called for a per-connection `cache_level` controlling the proxy's federated-credential cache granularity. The first cut stored it as a 3-value enum on the connection. But nothing app-side *reads* the value — it's pure pass-through to the proxy — and the only two cases that exist map exactly onto whether the connection has an `owner`. So instead of storing it (plus owning a UI selector, edit-preservation, and a migration), it's **derived at the proxy-exposure boundary**:

| connection | derived `cache_level` | why |
|---|---|---|
| **unowned** (Source-managed shared, ~99% of products) | `system` | one cached credential, zero regression |
| **owned** (customer) | `product` | per-product, always safe |

This matches the `owner` field's existing documented semantics (*"absent = unowned (e.g. Source Cooperative-managed)"*).

## Changes (3 files, +62/-4)

- `data-connection.ts` — `DataConnectionCacheLevel` enum (the proxy wire vocabulary); `cache_level` is **optional/wire-only** on `S3WebIdentityRoleAuthenticationSchema` and never stored.
- `sanitize-data-connection.ts` — `withDerivedCacheLevel` attaches the derived value to the federated `authentication` the proxy reads; no-op for every other auth type.
- `sanitize-data-connection.test.ts` — unowned ⇒ `system`, owned ⇒ `product`.

## Notes
- The proxy contract is unchanged — it still receives `cache_level` in the federated config, just derived instead of stored.
- `account` level is reserved for a future customer needing account-level caching with an account-scoped trust policy; nothing emits it yet (YAGNI).
- No admin-UI selector and no model changes — so #328 sheds the `cache_level` selector it would otherwise owe.
- **Even-lazier alternative not taken:** the proxy already receives `owner`, so it could derive this itself with zero app code. Kept it app-side so the policy lives in one place and the proxy stays a dumb cache.

## Verification
- `npx tsc --noEmit` — clean
- `jest` data-connection types + actions + sanitize + api routes — 60 passing

### DIFF
--- a/src/lib/api/sanitize-data-connection.test.ts
+++ b/src/lib/api/sanitize-data-connection.test.ts
@@ -93,14 +93,27 @@ describe("sanitizeDataConnection", () => {
     ).toBeUndefined();
   });
 
-  test("keeps secret-less auth when the caller may view the config", () => {
+  test("keeps secret-less auth (with derived cache_level) when viewable", () => {
     const c = conn({
       allowed_visibilities: [ProductVisibility.Public],
       authentication: roleAuth as never,
     });
+    // Unowned ⇒ Source-managed shared connection ⇒ system-level caching.
     expect(
       sanitizeDataConnection(c, sessions["anonymous"]).authentication
-    ).toEqual(roleAuth);
+    ).toEqual({ ...roleAuth, cache_level: "system" });
+  });
+
+  test("derives cache_level=product for an owned federated connection", () => {
+    const c = conn({
+      allowed_visibilities: [ProductVisibility.Restricted],
+      owner: "acme",
+      authentication: roleAuth as never,
+    });
+    expect(sanitizeDataConnection(c, memberOfAcme).authentication).toEqual({
+      ...roleAuth,
+      cache_level: "product",
+    });
   });
 
   test("strips secret-less auth when the caller may NOT view the config", () => {
--- a/src/lib/api/sanitize-data-connection.ts
+++ b/src/lib/api/sanitize-data-connection.ts
@@ -1,12 +1,37 @@
 import {
   Actions,
   DataConnection,
+  DataConnectionAuthenticationType,
+  DataConnectionCacheLevel,
   DataConnectionObjectSchema,
   DataConnectionSchema,
   isSecretBearingAuth,
   UserSession,
 } from "@/types";
 import { canViewDataConnectionConfig, isAuthorized } from "./authz";
+
+/**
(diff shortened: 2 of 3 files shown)


alukach:
Closing — superseded by a simpler design (no app change needed).

Cache granularity is a **proxy** concern, not an app schema field. The proxy already receives `owner` in the secret-less connection config (`sanitizeDataConnection` never strips it), so it keys its federated-credential cache directly:

- **owned connection** ⇒ cache per product-grained subject `scv1:conn:{conn}:{account}/{product}` — every unique (account, product) does its own `AssumeRoleWithWebIdentity`, so the customer's IAM **trust policy is the enforcement boundary** for which accounts/products may use the connection.
- **unowned** (Source-managed shared, ~99% of products) ⇒ cache per connection — one credential reused across all products (zero regression).

So there's no `cache_level` to store or even put on the wire. The `S3WebIdentityRole` schema (#332) is already complete; #326 needs no further app change. The cache rule lands proxy-side (data.source.coop #141).

PR #384, opened 2026-06-24
feat: OAuth2 device authorization verification UI

alukach:
## What & why

Ory Hydra ships **no hosted UI** for the OAuth2 Device Authorization Grant (RFC 8628) — Account Experience covers only login/registration/consent. Without a verification UI, `/oauth2/device/verify` errors with *"configuration key urls.device.verification is not set"*, so the `source-coop` CLI's `--flow device-code` can't complete. This PR provides that UI.

## Changes

Three files under a new top-level `/device` route — no new dependencies:

| File | Role |
|---|---|
| `src/app/device/page.tsx` | Verification page: reads `device_challenge` (and optional `user_code` to pre-fill), renders the code-entry form |
| `src/app/device/actions.ts` | Server action: `PUT /admin/oauth2/auth/requests/device/accept`, origin-checks `redirect_to`, hands off to the normal login/consent flow |
| `src/app/device/success/page.tsx` | Static "device authorized" landing page (`urls.device.success`) |

### Notes
- **Top-level route** so it inherits the root `<Theme>` but skips the `(app)` group's nav/footer/verification-banner chrome (which assumes an authenticated session — wrong for a pre-auth interstitial).
- **Pure Server Components**: the form posts straight to the server action; bad/expired codes bounce back to `/device?error=…`, so no client-side state.
- Mirrors the existing **raw-`fetch`-to-Hydra-admin** pattern in `lib/actions/proxy-credentials.ts` (same admin endpoint shape, `assertHydraOrigin`-style SSRF check). Admin key stays server-side and is already a registered sensitive value in `LOGGER`.

## Out of scope (ops wiring)

Point the Ory project config at these pages (workspace auth required):
```
urls.device.verification -> https://<site>/device
urls.device.success      -> https://<site>/device/success
```
The OAuth2 client must also have the `urn:ietf:params:oauth:grant-type:device_code` grant (already enabled on the staging `source-coop-cli` client).

## Verify
- `npm run type-check` — clean
- `npm run lint` — clean
- Manual: run `dev`, set the two Ory URLs to `http://localhost:3000/device[/success]`, then run the CLI device flow end-to-end.

One thing to confirm on first real run: Hydra is assumed to pass the challenge as `?device_challenge=` (matches `login_challenge`/`consent_challenge` and the Ory sample). If the page reports a missing challenge, adjust the param read in `page.tsx`.

### DIFF
--- a/src/app/device/actions.ts
+++ b/src/app/device/actions.ts
@@ -1,0 +1,82 @@
+"use server";
+
+import { redirect } from "next/navigation";
+import { CONFIG } from "@/lib/config";
+import { LOGGER } from "@/lib/logging";
+
+/**
+ * Server action backing the device-verification form (RFC 8628).
+ *
+ * Ory Hydra has no hosted UI for the device flow, so this app provides one:
+ * Hydra redirects the user's browser to `urls.device.verification` (our
+ * `/device` page) with a `device_challenge`, the user enters the `user_code`
+ * shown in their terminal, and we accept it via Hydra's admin API. Hydra then
+ * returns a `redirect_to` that runs the normal login/consent flow (reusing the
+ * Account Experience UIs) before landing on `urls.device.success`.
+ *
+ * On a bad/expired code we bounce back to `/device` with an `error` marker so
+ * the page can render a message — keeping the page a pure Server Component with
+ * no client-side state.
+ */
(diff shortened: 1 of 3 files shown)


alukach:
Not supported by Ory, see https://github.com/source-cooperative/source-coop-cli/pull/5#issuecomment-4806323160

Issue #288, opened 2026-04-11
[Proposed Feature] improved data preparation for browser upload data


alukach:
**Description of Feature:**

When a user uses the drag and drop uploader on the website, we should help them conform to best practices around file names and avoiding uploading litter. For example, this may be running the file names through a kabob case formatter and ignoring `.DS_Store` files. Additionally, we may want to auto assign mime types to the uploaded files based on the file names (see https://github.com/source-cooperative/source.coop/issues/198). The appropriate workflow might involve an intermediary screen that demonstrates what the uploaded files will be once they are uploaded, allowing users to verify the formatting that we apply. We could allow users to directly opt out of these features or simply instruct them to use direct command line access to bypass these features.

**What value is this feature adding to Source Cooperative?**

Help maintain best practices for users of the web uploader, which is where we would expect the majority of less experienced users to be interacting with the system.

PR #406, opened 2026-07-01
fix(uploader): reject filenames with characters that break upload signature

### LINKED ISSUE #405: [Bug] "Error: signature mismatch" when uploading file with parenthesis in filename
**Description of Bug:**

Upload fails when the filename contains a parenthesis. 

**Steps to Reproduce:**

- Step 1 - Enter edit mode for a data product you own
- Step 2 - Upload a filename that contains a parenthesis (like "test().txt")

**Expected Behavior:**

- If parenthesis characters are not allowed, an actionable error should be returned.
- If parenthesis characters are allowed, the upload should complete successfully

**Actual Behavior:**

An error is displayed: "Error: signature mismatch"

**Screenshots:**

Image

**Additional Context:**

tylere:
## Summary
- Uploading a filename containing `!`, `'`, `(`, `)`, or `*` fails with a cryptic "Error: signature mismatch". The storage proxy's SigV4 signature verification doesn't handle those characters the way S3 does (an external service, outside this repo, so the signing bug itself can't be fixed here).
- This implements the issue's other acceptable behavior instead: reject those filenames up front with a clear, actionable error rather than failing deep inside the upload.
- `UploadQueueManager.process()` now marks any queued item whose key contains those characters as `error` with an explanatory message. Both upload entry points (drag-and-drop and the file picker) funnel through this same queue, so the check applies uniformly.

## Test plan
- [x] Added `upload-queue-manager.test.ts` covering the rejection and the unaffected happy path.
- [ ] Could not run `npm test` / `npm run lint` in this sandbox (no `node_modules`, no permission to install dependencies) - please run before merging.

Closes #405

### DIFF
--- a/src/components/features/products/object-browser/DirectoryRow.tsx
+++ b/src/components/features/products/object-browser/DirectoryRow.tsx
@@ -68,6 +68,7 @@ export function DirectoryRow({
   const {
     cancelUpload,
     retryUpload,
+    removeUpload,
     getUploadsForScope,
     deleteObject,
     deletePrefix,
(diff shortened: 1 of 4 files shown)


tylere:
Space characters also seem to cause the error:

Image

Also it is not clear how to remove the messages for the uploads that failed, even after a subsequent successful upload. Do the messages eventually disappear?

tylere:
@claude incorporate the last comment

alukach (commented):
There's some thought in us guiding towards uploads-best-practices on the frontend (#288) but I think this is a legitimate bug that should be fixed on data.source.coop

---

UPDATE: This has been fixed by https://github.com/source-cooperative/data.source.coop/pull/176

Issue #324, opened 2026-06-01
[Proposed Feature] Viewer for Zarr


tylere:
**Description of Feature:**

Create a default viewer for Zarr datasets.

Creating a generic viewer may be challenging, in part because Zarr have have many dimensions, and chunking choices may strongly impact how responsive a viewer can be (along various dimensions). 

**What value is this feature adding to Source Cooperative?**

There are many data products that include Zarr files. A viewer would be helpful in determining what dimensions exist and the spatial and temporal intervals of the dataset.

tylere:
It would be beneficial to indicate what Zarr version is being used, and to encourage usage of the most recent version (v3).

tylere:
Zarr viewer development: https://github.com/source-cooperative/zarr-viewer
Deployed viewer: https://source-cooperative.github.io/zarr-viewer/

PR #415, opened 2026-07-03
feat: zarr/icechunk viewer

tylere:
## What I'm changing

Add a viewer for Zarr and Icechunk directories.

## How I did it

Added a new viewer that works on on .zarr and .icechunk directories.

## How you can test it

- [ ] Navigate to Zarr or Icechunk directories in Source. Verify that an iframe with the viewer appears and renders the dataset.
  - [ ] https://source-cooperative-git-feat-zarr-icechunk-viewer-radiantearth.vercel.app/bkr/silam-dust/silam_aerosol.icechunk
  - [ ] https://source-cooperative-git-feat-zarr-icechunk-viewer-radiantearth.vercel.app/dynamical/noaa-hrrr-forecast-48-hour-virtual/v0.5.0.icechunk
  - [ ] https://source-cooperative-git-feat-zarr-icechunk-viewer-radiantearth.vercel.app/bkr/geos/geos_v2_15min.icechunk
- [ ] Navigate to a Tif file, and verify that its viewer still works.


alukach:
@tylere could you rebase this off of `main` and perhaps provide a few links to some existing zarr and icechunk stores on staging?

tylere:
Fixes https://github.com/source-cooperative/source.coop/issues/324

tylere:
Added examples to the issue description.

tylere:
Addressed both review findings in 6494159.

**1. Bare Zarr v2 root array (`probeZarr`)** — Added a "Tier 1/2d" check to `probeZarr`: after the `.zgroup` group case, it GETs the root `.zarray` and, if it's a valid v2 array, renders it (`path: ""`). This restores parity with the v3 path (`findArrayV3`'s "root itself is an array" case). Added two reproducing tests to `probe.test.ts` covering the bare-root-array layout (chunk canary `ok` and `inconclusive`) — red before the fix, green after.

**2. Redundant `getExtension` (page.tsx)** — Added an `isStoreExtension(ext)` type guard in `files.ts`, reused it inside `isViewableStorePath`, and changed the store block to gate on the already-computed `storeExtension` (`if (isStoreExtension(storeExtension))`) instead of recomputing the extension via `isViewableStorePath`. The `ext is string` predicate also preserves the narrowing `StorePreview`'s `extension: string` prop needs.

Verification: reproducing tests red → green, typecheck clean, full suite passes (469 tests), no new lint issues.

_Separately, this branch also now recognizes the Icechunk spec 2.x on-disk layout (root `repo` info object, no `refs/` prefix) so the `.icechunk` examples above render — the probe previously only understood the legacy `refs/branch.*/ref.json` layout._

mrshll:
subscribed ❤️

alukach (dismissed):
In #421, we broke the object previews out from the object list card to allow them to take the entirety of the width of the app container:

image


However, this PR places the Zarr/Icechunk previews within the object list card which confines their horizontal width:

image

These previews should be instead be in the object preview card

alukach (approved):

mrshll:
!!! I can't wait to play with this

PR #418, opened 2026-07-04
feat: add external link icon to open data viewers in a new tab

### LINKED ISSUE #417: [Proposed Feature] External link icon for data viewers
**Description of Feature:**

If the object or prefix displays a data viewer, also show an external link icon that will open the data viewer in a separate tab/window.

Some of the data viewers are hard to use when the dimensions are constrained by the iframe of the source.coop website, and it would be useful show the same content in a full browser window.

tylere:
## Summary

Adds an "Open in new tab" link with an external-link icon above embedded data viewer iframes (pmtiles, parquet, CSV/TSV, COG, images, PDF, 3D models, zip, JSON), so viewers that are hard to use inside the constrained iframe can be opened full-size in their own browser tab/window.

Closes #417

### DIFF
--- a/src/components/features/products/object-browser/ObjectPreviewExternal.tsx
+++ b/src/components/features/products/object-browser/ObjectPreviewExternal.tsx
@@ -2,7 +2,8 @@ import "server-only";
 
 import { LOGGER } from "@/lib";
 import { fileSourceUrl } from "@/lib/urls";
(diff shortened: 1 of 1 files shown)


tylere:
@alukach requesting your review on this small change, because it changes the UI slightly.

alukach (approved):
Great addition!

Issue #258, opened 2026-03-30
[Proposed Feature] Analytics - Storage


alukach:
**Description of Feature:**

Data providers should be able to see the volume of data they have stored both within and across products within Source Cooperative. Ideally, they should be able to see a trend of how this volume data changed.

**What value is this feature adding to Source Cooperative?**

This seems like it would be generally useful information, both for product owners and as system administrators. Particularly if we need to charge accounts for storage.

LanesGood:
Do organizations currently have a private provider view?

In terms of both org-wide and product-centric analytics on both storage and usage, another good comparison could be the "Insights" tab on Github, with easily accessible charts and figures. What seems useful here is also the adjustable time period, and from a dev perspective it seems like a lot of this is calculated on the fly, rather than needing to store this information in a database (at least that's my assumption based on the fact that queries seem to need to be first run when you visit the page).

Does Amazon also provide this info for S3 in an easily digestible way? Not looking to replicate or provide feature parity, but could be interesting to see what is common/expected in the space and what is already visible for users who may be using other cloud storage providers.

alukach:
> Do organizations currently have a private provider view?

Not exactly. It's more akin to the style of Github, where anyone can view github.com/source-cooperative but if you're authenticated and authorized, you'll see things not intended for the public such as private repos, buttons to access settings, etc.

However, I do think it's reasonable to assume that someday, the homepage of source.coop will be more of a dashboard "see all the things" view when a user is authenticated. However, that's probably scope creep for this exercise.

> In terms of both org-wide and product-centric analytics on both storage and usage, another good comparison could be the "Insights" tab on Github, with easily accessible charts and figures.

I think this is a very good suggestion.

> What seems useful here is also the adjustable time period, and from a dev perspective it seems like a lot of this is calculated on the fly, rather than needing to store this information in a database

The [Workers Analytics Engine](https://developers.cloudflare.com/analytics/analytics-engine/) seems to be very capable of querying time spans on the fly. However, I think that's probably less important for this Storage Analytics feature than it is for the Usage Analytics feature (#257). 

> Does Amazon also provide this info for S3 in an easily digestible way? Not looking to replicate or provide feature parity, but could be interesting to see what is common/expected in the space and what is already visible for users who may be using other cloud storage providers.

Yes, within an S3 Bucket, you can click on the "Metrics" tab:

[...]

PR #421, opened 2026-07-06
feat: analytics suite — product/object usage cards + admin traffic explorer

### LINKED ISSUE #257: [Proposed Feature] Analytics - Usage
## Description of Feature

The data proxy is currently collecting analytics data via [Worker Analytics Engine](https://developers.cloudflare.com/analytics/analytics-engine/). We can surface this data to end-users.

For every request, we store:
* `account_id`
* `product_id`
* `file_path`
* `method`
* `user_id` *(currently unused)*
* `country`
* `content_type`
* `bytes_sent`
* `status_code`

Data is retained for 90 days.

> [!NOTE]
> Analytics data can be sampled on both write & read, meaning that the values are close to but exactly match reality.

## What value is this feature adding to Source Cooperative?

These statistics empower us to demonstrate access summaries on system, account, product, prefix, and file level.

We can display access data (e.g. download counts, byte volume) over time and/or client location.

This could be useful for data consumers to determine the popularity of a product against others. This could be useful for data providers to help understand how many people are using their products.

alukach:
Analytics for source.coop, backed by the Cloudflare Analytics Engine dataset the data proxy writes. Closes #257.

## Three surfaces

**Product Analytics card** (product page, right column)
- Public: Downloads / Data served / Countries over a daily downloads chart (hover a bar for that day's numbers).
- Owners/maintainers/admins additionally get a USERS tab (unique IPs, registered users, anon requests, per-IP download-frequency histogram) and a "View all analytics →" link.
- Streams via Suspense with a skeleton; hides itself entirely if `CF_ANALYTICS_*` is unset or queries fail.

**Product analytics page** — `/{account}/{product}?tab=analytics` (manager-only, 404 otherwise)
- 7d/30d/90d windows, stats row with daily average, downloads chart beside a by-country ranking, top-files table (per-file downloads + data served), same USERS tab.
- A PRODUCT | ANALYTICS strip appears at the product root for managers.
- Routing: middleware rewrites the query-param URL to an internal `/-/analytics` route (layouts can't read search params; the `-` segment can't shadow real object paths). See the routing-review comment for the full analysis.

**Admin explorer** — `/admin/analytics` (admin tool)
- URL-driven: windows 24h/72h/1wk/1mo/3mo, account/product filters, composable group-bys (account, product, country, IP hash) → stacked bar timeseries with bytes⇄requests toggle + ranked totals table. Country codes expand to names ("United States (US)").

## Layout rework (product page)

Two columns — summary + Product Contents left; Product Details + Product Analytics stacked right — with the README and a new titled "Object Preview" card (via an `@preview` parallel slot) as full-width rows below.

## Data layer (`src/lib/clients/analytics/`)

- Server-only AE SQL API client; sampling-correct aggregates (`_sample_interval`-weighted), `COUNT(DISTINCT …)` uniques.
- Windows are day-aligned in SQL (`toStartOfDay(NOW() - INTERVAL 'N-1' DAY)`) so headline, bars, and breakdowns cover identical spans.
- Served filter `blob4 = 'GET' AND double2 IN (200.0, 206.0) AND blob2 != ''` (float literals — AE's validator rejects Double-vs-Integer comparisons; HEAD/listing traffic excluded).
- Escaped literals + whitelisted windows/columns (AE has no parameterized queries); bounded top-N + "Other" derivation for unbounded group cardinality.
- `unstable_cache`: 4 h for card/page stats, 15 min for the admin explorer.

## Config

`CF_ANALYTICS_ACCOUNT_ID`, `CF_ANALYTICS_API_TOKEN` (Account Analytics:Read), `CF_ANALYTICS_DATASET` — see `.env.example`. Verified working against the production dataset on the preview deployment.

## Deferred

Storage metrics (needs Metadata Catalog), account-profile analytics, pie charts, per-object analytics (data layer supports it; UI removed), blob9 range display.

## Reviews & testing

Multi-agent review rounds are summarized in the PR comments (correctness/security/React pass; routing-technique pass) — all confirmed findings fixed. 48 suites / 403 tests, including SQL escaping/injection, sampling math, window alignment, top-N/Other math, and middleware rewrite tests.

> [!NOTE]
> The PR preview is pointed at **production analytics** data so it may reference products that don't exist on staging. However, I think this displays more interesting data than the staging data does. If we merge this, I will point the staging/preview branches to the staging analytics data.


alukach:
## Review pass (ponytail + 3 independent agents)

Ran an over-engineering review plus three parallel review agents (correctness/query math, security/robustness, React/Next/Recharts). Fixed in b95bd1b:

| # | Severity | Finding | Fix |
|---|----------|---------|-----|
| 1 | High | Product-card hover never fired — recharts 3 passes `activeTooltipIndex` as a numeric **string**; the `typeof === "number"` guard rejected every event (confirmed against installed recharts internals by two agents independently) | `parseActiveIndex()` accepts string/number, bounds-checked, unit-tested |
| 2 | Med | `NOW() - 28 DAY` touches 29 UTC days: the partial oldest day counted in totals but appeared in no bar | window is now `>= toStartOfDay(NOW() - 27 DAY)`, identical span for series + totals |
| 3 | Med | Requests-only "Other" remainder (zero-byte 200s) was gated on bytes and vanished from the Requests view and table sums | gate on either metric; regression test |
| 4 | Med | No timeout on the AE fetch — a hung API would hold product-page Suspense boundaries (and invocations) open | `withTimeout(…, 15s)` per repo convention |
| 5 | Med-low | `toLocaleString()` in SSR'd client components → hydration mismatch for non-en locales | explicit `Intl.NumberFormat("en-US")` |
| 6 | Low | `key in WHITELIST` matches prototype keys (`?window=constructor` → malformed SQL, admin-only broken page) | `Object.hasOwn` |
| 7 | Low | A11y/API nits: inert tab stop on the product chart, `aria-label` on a generic div, deprecated `align` on Radix table cells | `accessibilityLayer={false}` + `role="img"`, `justify="end"` |

**Reviewed and accepted as-is:** 12-char client-hash display keys could theoretically collide (48-bit prefix among ≤25 groups — negligible); first-view cache-key fan-out on public pages (2 AE queries per unique product/object per hour — bounded by the 1h cache, now also timeout-capped); SQL escaping confirmed sound by dedicated injection review (all values quoted, structure whitelisted, backslash-then-quote ordering correct).

**Also checked:** repo coverage threshold (80%) isn't enforced — CI runs `npm run test` without `--coverage`, and repo-wide coverage is ~34% pre-existing.

Post-fix: 394 tests pass, type-check + lint clean.

alukach:
## 3-agent review of the ?tab=analytics → /-/analytics routing technique

**Verdict: sound-with-caveats (unanimous; no restructuring).** The architect agent ruled out every alternative against the constraints — conditional render in the page (renders inside the contents card), layout tab-switching (layouts can't read searchParams and don't re-render on query-only navigation), parallel-slot swapping (path-driven), intercepting routes (overlay semantics), moving the shell into the page (dismantles the @readme/@preview slot architecture and the layout's persistence purpose). The only simpler design is making `/-/analytics` the visible URL and dropping the rewrite — worth remembering if the `?tab=` URL ever becomes negotiable.

Findings fixed in af89eab5:

| Severity | Finding | Fix |
|---|---|---|
| Med | Middleware matcher exclusions were unanchored prefixes — accounts starting with `api`/`logo`/`favicon` (legal per ID_REGEX) never hit middleware: dead ANALYTICS tab, skipped Ory handling | anchored to `api/`, `logo/`, `favicon/`, `favicon.ico` |
| Low-med | Rewrite matched ANY two-segment path and ran before the Ory branch — `/sessions/whoami?tab=analytics` skipped Ory; `/admin/analytics?tab=analytics` 404'd | Ory branch runs first; rewrite skips known top-level segments |
| Low | `generateMetadata` on `/-/analytics` gated only product-read → non-managers of public products could get the "— Analytics" title on the 404 | full PutRepository gate in metadata too; unauthorized ≡ nonexistent |
| Low | `/admin/analytics` relied solely on the admin layout gate (layouts aren't an auth boundary per Next docs) | in-page `isAdmin` → `notFound()` |
| UX | No `loading.tsx` on `/-/analytics` — tab click froze on the old view until both queries resolved | skeleton added |
| Robustness | The rewrite was untested, silent-failure indirection | `handleProductAnalyticsTab` exported + 6 unit tests |

Verified clean by the reviewers: route precedence via the build manifest (`/-/analytics` beats the catch-all; only the single object key `-/analytics` is shadowed, and its bytes stay downloadable via the direct data URL); trailing-slash normalization precedes middleware; RSC/prefetch requests get rewritten correctly with distinct router-cache keys; `unstable_cache` entries are product-scoped and viewer-independent; the route renders dynamically (session cookies) so no full-route-cache cross-viewer leak; users-tab data provably never enters the public RSC payload; pathname traversal can't escape the rewrite target shape.

tylere:
From a statistical standpoint, I think we should use multiples of 7-day windows for analytics, to avoid capturing day-of-the-week patterns. (e.g. a 30-day window will sometimes be 22 weekdays + 8 weekend days and other times will be 20 weekdays + 10 weekend days).
So instead of 7d/30d/90d use 7d/28d/91d windows.

tylere:
Some of the product links on the [admin/analytics](https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/admin/analytics) page are broken.
Examples:
- [tge-labs/aef](https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/tge-labs/aef)
- [ftw/global-data](https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/ftw/global-data)

tylere:
On the Admin Analytics page, it would be useful to be able to move forward and backward by the time range shown. (e.g. When viewing "1 week", be able to shift forward or backward by a week)

tylere:
The product metrics displayed don't currently match the [V1 Proposal](https://github.com/source-cooperative/source.coop/issues/257#issuecomment-4804710624). Is that gated on Lane's review and update of the mocks?

tylere:
Overall, it is exciting to see this preview!

alukach:
@tylere thanks for the review. my replies:

> From a statistical standpoint, I think we should use multiples of 7-day windows for analytics, to avoid capturing day-of-the-week patterns. (e.g. a 30-day window will sometimes be 22 weekdays + 8 weekend days and other times will be 20 weekdays + 10 weekend days).
> So instead of 7d/30d/90d use 7d/28d/91d windows.

This seems reasonable. Will implement. The only downside is that the 28d window has fewer check marks in the chart which makes it look a bit less rich.

> Some of the product links on the [admin/analytics](https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/admin/analytics) page are broken. Examples:
> 
> * [tge-labs/aef](https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/tge-labs/aef)
> * [ftw/global-data](https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/ftw/global-data)

I tried to call this out in the NOTE at the bottom of the PR description but this is a side effect of using production analytics on the staging side. 

> On the Admin Analytics page, it would be useful to be able to move forward and backward by the time range shown. (e.g. When viewing "1 week", be able to shift forward or backward by a week)

Yeah, perhaps this should just be a free date selector input with some common presets (eg last 30 days)

> The product metrics displayed don't currently match the [V1 Proposal](https://github.com/source-cooperative/source.coop/issues/257#issuecomment-4804710624). Is that gated on Lane's review and update of the mocks?

How so? Can you be more specific?

tylere:
> > The product metrics displayed don't currently match the [V1 Proposal](https://github.com/source-cooperative/source.coop/issues/257#issuecomment-4804710624). Is that gated on Lane's review and update of the mocks?
> 
> How so? Can you be more specific?

It is missing (presumably because the source is different):
- storage - total objects
- storage - total bytes

The "view all analytics" page for a product wasn't covered in the V1 proposal (although the link does appear in a screenshot). I'd like to hear Lane's thoughts on what might appear on it (and whether it should be included in V1).

Michelle was going to try to review V1 tonight. If she doesn't have time, I think we probably should move toward getting new mocks from Lane (or just get his input on this PR).

alukach:
> It is missing (presumably because the source is different):
> 
> * storage - total objects
> * storage - total bytes

Ah, those are _storage_ metrics (#258). This PR aims to address _usage_ analytics (#257). While I do think that we may reasonably want to someday display storage analytics alongside the usage metrics, **I recommend that we save that for a separate feature at a later date.**

> The "view all analytics" page for a product wasn't covered in the V1 proposal (although the link does appear in a screenshot). I'd like to hear Lane's thoughts on what might appear on it (and whether it should be included in V1).

The mocks describe that the "view all analytics" link should take an admin/owner to the analytics tab.

image

Which is then described further in the mockups as:

image

So I generally ran with that. Perhaps I misunderstood the intention of the V1 proposal. I read through it and interpreted it to be codifying the concepts present in the design mocks. 

I guess a grounding question to this is: **what questions do we want a product owner or product viewer to be able to answer that they can't currently answer in the design as it stands (within this branch or in the mocks) today?**

tylere:
> So I generally ran with that. Perhaps I misunderstood the intention of the V1 proposal. I read through it and interpreted it to be codifying the concepts present in the design mocks.

My intention with the V1 proposal was to present a minimum set of metrics and how they would be shown on the website (pulled from a subset of the design mocks). If other metrics also make it into V1, I am fine with that.

I do think that storage metrics should be in V1 because they are something that the product owners should be aware of (even though they are not currently being charged based on storage). I think that the ability to calculate `(bytes served) / (bytes stored)` is important. Storage metrics could be addressed on a separate PR, but if both storage and usage metrics are in the Product Analytics box, it may be easier to reason about them in a single PR.

tylere:
I managed to get to a state where the bar chart is very faint...
https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/admin/analytics?groupBy=product&from=2026-04-15&to=2026-07-14&interval=360
image

alukach:
> I managed to get to a state where the bar chart is very faint... https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/admin/analytics?groupBy=product&from=2026-04-15&to=2026-07-14&interval=360 image

Resolved now: https://source-cooperative-git-worktree-analytics-suite-radiantearth.vercel.app/admin/analytics?groupBy=product&from=2026-04-16&to=2026-07-15&interval=360

tylere:
Minor nit: the current ordering of the "Downloads" metrics on the Product Analytics tab is ordered as:

- **\# of downloads** ("DOWNLOADS")
- volume of data ("DATA SERVED")
- \# of countries ("COUNTRIES")
- **\# of downloads** per day  ("DAILY AVG")

image

I think it would be easier to interpret if the two **# of download** metrics were together:

- **\# of downloads** ("DOWNLOADS")
- **\# of downloads** per day  ("DAILY AVG")
- volume of data ("DATA SERVED")
- \# of countries ("COUNTRIES")

tylere:
I wonder if it be worth adding "user-agent" as a dimension to group/filter by? I realize it could be noisy, and users can spoof it, but it may yield some insights into the types of clients that are being used to pull data.

tylere (approved):

PR #430, opened 2026-07-15
fix(product): shorten product page card titles

alukach:
## What I'm changing

When working on #421, I noticed that almost every card (save for "README") began with either "Product" or "Object": 

| image |
|:--:|
| _"Product ..." overload_ |

This feels unnecessary and overly noise.

## How I did it

Renames the product page cards: "Product Contents" → "Contents" and "Product Details" → "Details", including the loading skeleton. The cards live on the product page, so the prefix was redundant.

Left "Product Title" in the creation form unchanged — it is a form field label, not a card title.

| image |
|:--:|
| _Before_ |

| image |
|:--:|
| _After_ |

### DIFF
--- a/src/app/(app)/[account_id]/[product_id]/(product)/[[...path]]/layout.tsx
+++ b/src/app/(app)/[account_id]/[product_id]/(product)/[[...path]]/layout.tsx
@@ -92,7 +92,7 @@ export default async function ProductLayout({
         <Dropzone product={product} prefix={prefix}>
           <Card>
             <SectionHeader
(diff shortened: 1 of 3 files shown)


tylere (approved):
That makes sense to me... to avoid superfluously, gratuitously, excessively, needlessly, unnecessarily, repetitively, recurrently, iteratively, redundantly, tautologically, and pleonastically pleonastic labeling.

LanesGood (approved):
Agreed, nice catch.

PR #452, opened 2026-07-22
fix(data-connections): target GCS, align DataProvider with storage_type

alukach:
## Problem
Creating a GCP data connection persisted `details.provider: "gcp"`, but the backend (data.source.coop) addresses Google Cloud Storage as `"gcs"`. The same latent mismatch existed for Azure (`"az"` vs `"azure"`).

Root cause: `DataProvider` (connection config) and `ProductMirror.storage_type` (a per-mirror field) were **two vocabularies** bridged by a hand-written `STORAGE_TYPE_BY_PROVIDER` map **duplicated** in `products.ts` and `product-mirrors.ts`.

### 1. Align `DataProvider` values to the backend vocabulary
```
enum DataProvider { S3 = "s3", Azure = "azure" /* was "az" */, GCS = "gcs" /* was GCP="gcp" */ }
```
Both translation maps deleted. Member `GCP`→`GCS`, schema/type `Gcp*`→`Gcs*`. Auth type stays `gcp_workload_identity` (WIF is a GCP-*platform* feature, not storage). Form label → **"Google Cloud Storage"**.

### 2. Remove `ProductMirror.storage_type` entirely
Nothing branched on it — the proxy keys off `details.provider`, and the only source.coop consumer was a display label. It was a denormalized copy of the provider. Dropped the field; the mirror "Type" label now derives from the connection's provider. This dissolves the "two vocabularies" problem rather than just aligning it, and removes both provider→storage_type casts.

### 3. Migration (legacy `az`/`gcp` connection rows)
DynamoDB reads bypass Zod, so `normalizeConnection` in the data-connections client remaps `"az"→"azure"` / `"gcp"→"gcs"` on `fetchById`/`listAll`. Marked `ponytail:` — drop after backfill. Removing `storage_type` needs **no** migration: `ProductMirrorSchema` is a plain `z.object`, so old records still carrying the field parse fine (unknown key stripped).

## Backend compatibility (verified against data.source.coop)
- **S3** unchanged. **Azure** accepts both `"az"` and `"azure"` → compatible during rollout. **GCS**: old `"gcp"` never worked (hit the error arm); `"gcs"` is accepted. So this is the fix, not a regression.
- ⚠️ Separate prerequisite for GCS to actually connect: the data proxy must enable the multistore `gcp` cargo feature — **and** that currently fails to compile for `wasm32-unknown-unknown` (object_store's GCS credential signing uses `ring::SystemRandom`, unsupported on that target). Tracked separately; not in this PR.

## Verification
`tsc --noEmit` clean · `jest` affected suites **116/116** · lint clean on changed files.

### DIFF
--- a/src/app/(app)/edit/product/[account_id]/[product_id]/data-connections/page.tsx
+++ b/src/app/(app)/edit/product/[account_id]/[product_id]/data-connections/page.tsx
@@ -56,6 +56,7 @@ export default async function ProductDataConnectionsPage({
       {
         name: c.name,
         bucket: "bucket" in c.details ? c.details.bucket : c.details.container_name,
+        provider: c.details.provider,
       },
     ])
   );
--- a/src/components/features/data-connections/DataConnectionForm.tsx
+++ b/src/components/features/data-connections/DataConnectionForm.tsx
@@ -28,11 +28,11 @@ interface DataConnectionFormProps {
   ownerAccountId?: string;
 }
 
-// Storage providers limited to those with a `details` schema (S3, Azure, GCP).
+// Storage providers limited to those with a `details` schema (S3, Azure, GCS).
 const providerOptions: Array<{ value: DataProvider; label: string }> = [
   { value: DataProvider.S3, label: "S3 / S3-compatible (R2, MinIO)" },
   { value: DataProvider.Azure, label: "Azure" },
-  { value: DataProvider.GCP, label: "GCP (GCS)" },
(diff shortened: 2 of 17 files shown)


alukach:
Split into two focused PRs:
- #453 — fix(data-connections): target GCS, align DataProvider values
- #454 — refactor(mirrors): drop denormalized ProductMirror.storage_type (stacked on #453)

Closing this combined PR in favor of those.

PR #453, opened 2026-07-22
fix(data-connections): target GCS, align DataProvider with storage_type

alukach:
## Problem
Creating a GCP data connection persisted `details.provider: "gcp"`, but the backend (data.source.coop) addresses Google Cloud Storage as `"gcs"`. The same latent mismatch existed for Azure (`"az"` vs `"azure"`).

Root cause: `DataProvider` and a mirror's storage type were two vocabularies bridged by a hand-written `STORAGE_TYPE_BY_PROVIDER` map **duplicated** in `products.ts` and `product-mirrors.ts`.

## Change — align `DataProvider` values to the backend vocabulary
```
enum DataProvider { S3 = "s3", Azure = "azure" /* was "az" */, GCS = "gcs" /* was GCP="gcp" */ }
```
- Both translation maps deleted; a mirror's storage type is just the connection's provider.
- Member `GCP`→`GCS`, schema/type `Gcp*`→`Gcs*`. Auth type stays `gcp_workload_identity` (WIF is a GCP-*platform* feature, not storage). Form label → **"Google Cloud Storage"**.

## Migration (legacy `az`/`gcp` rows)
DynamoDB reads bypass Zod, so `normalizeConnection` in the data-connections client remaps `"az"→"azure"` / `"gcp"→"gcs"` on `fetchById`/`listAll`. Marked `ponytail:` — drop after backfill.

## Backend compatibility (verified against data.source.coop)
- **S3** unchanged. **Azure** accepts both `"az"` and `"azure"` → compatible during rollout. **GCS**: old `"gcp"` never worked (hit the error arm); `"gcs"` is accepted. This is the fix, not a regression.
- ⚠️ Making GCS actually connect needs a separate backend change (enable the multistore `gcp` feature + a wasm-compatible signing path) — tracked in data.source.coop.

## Verification
`tsc` clean · `jest` affected suites pass · lint clean on changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #454, opened 2026-07-22
refactor(mirrors): drop denormalized ProductMirror.storage_type

alukach:
**Stacked on #453** (base is `fix/gcs-provider-alignment`) — review/merge that first; GitHub will retarget this to `main` once it lands.

## Why
Nothing branched on `storage_type`: the data proxy keys off the connection's `details.provider`, and the only source.coop consumer was a display label. It was a denormalized copy of the provider, bridged by a cast. This removes the field entirely — dissolving the "two vocabularies" problem rather than just aligning it.

## Changes
- Drop `storage_type` from `ProductMirrorSchema`; stop writing it in `createProduct` / `addProductMirror` (removes both provider→storage_type casts and the invariant test they needed).
- Mirror "Type" label now reads the provider from `connectionInfo` (sourced from the connection's `details.provider` on the edit page).
- **No migration:** `ProductMirrorSchema` is a plain `z.object`, so existing records still carrying `storage_type` parse fine (unknown key stripped).

## Verification
`tsc` clean · `jest` affected suites **116/116** · lint clean on changed files.

### DIFF
--- a/src/app/(app)/edit/product/[account_id]/[product_id]/data-connections/page.tsx
+++ b/src/app/(app)/edit/product/[account_id]/[product_id]/data-connections/page.tsx
@@ -56,6 +56,7 @@ export default async function ProductDataConnectionsPage({
       {
         name: c.name,
         bucket: "bucket" in c.details ? c.details.bucket : c.details.container_name,
+        provider: c.details.provider,
       },
     ])
   );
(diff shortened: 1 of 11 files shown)


alukach:
Re-split into https://github.com/source-cooperative/source.coop/pull/455 (value alignment) + https://github.com/source-cooperative/source.coop/pull/456 (storage_type removal, now also owns the map deletion).

PR #455, opened 2026-07-22
fix(data-connections): align provider values to the backend vocabulary

alukach:
## Problem
Creating a GCP data connection persisted `details.provider: "gcp"`, but the data proxy (data.source.coop) reads `details.provider` directly and only accepts `"gcs"`/`"gs"` for Google — never `"gcp"`. The same latent mismatch existed for Azure (`"az"` vs the accepted `"azure"`). The `provider→storage_type` map already emitted the right values, so mirrors were fine — **only the stored provider string was wrong**.

## Change — rename `DataProvider` values to what the backend accepts
```
enum DataProvider { S3 = "s3", Azure = "azure" /* was "az" */, GCS = "gcs" /* was GCP="gcp" */ }
```
- Member `GCP`→`GCS`, schema/type `Gcp*`→`Gcs*`. Auth type stays `gcp_workload_identity` (WIF is a GCP-*platform* feature, not storage). Provider labels spelled out in the form.
- **Maps kept** — the now-redundant `provider→storage_type` translation is left in place here and removed in the stacked follow-up (#456).

## Migration (legacy `az`/`gcp` rows)
A one-time script — `scripts/migrate-provider-values.ts` — converts persisted `details.provider` values (`"az"→"azure"`, `"gcp"→"gcs"`), matching the repo's existing migration convention (positional table arg, DynamoDB scan + guarded `UpdateCommand`, progress logging). **Dry-run by default; `--apply` to write.** Re-runnable (canonical rows are skipped). Run it (dry-run, then `--apply`) at/before deploy:
```
npx tsx scripts/migrate-provider-values.ts sc-prod-data-connections          # preview
npx tsx scripts/migrate-provider-values.ts sc-prod-data-connections --apply   # write
```
(Earlier revisions used a runtime read-shim; replaced with this script so the DB read path carries no legacy-value handling.)

## Backend compatibility (verified against data.source.coop)
**S3** unchanged. **Azure** accepts both `"az"` and `"azure"` → compatible during rollout. **GCS**: old `"gcp"` never worked (hit the error arm); `"gcs"` is accepted. Pairs with data.source.coop#191 (enables the GCS backend).

## Verification
`tsc` clean · `jest` affected suites green · lint clean · fixture `provider` fixed (`az`→`azure`) so module-load parsing passes.


alukach:
Fixed in de9b172c — `fixtures/data-connections.json` azure connection `provider` `az`→`azure`. Good catch; `utils.mock.ts` parses this fixture at module load so it took down every importing suite. Verified `authz.test.ts` (45) and the full fixture-consuming suites pass. Propagated the same fix into the stacked #456.

The remaining `DropdownSection.integration.test.tsx` failure is pre-existing on `main` and unrelated to this PR (doesn't touch that component).

alukach:
Replaced the runtime `normalizeConnection` read-shim with a one-time migration script (`scripts/migrate-provider-values.ts`, dry-run by default, `--apply` to write) per request — the DB read path no longer carries legacy-value handling. Propagated into the stacked #456.

Issue #471, opened 2026-08-03
Epic: mobile web performance — landing page FCP/LCP 6.1s on 4G


alukach:
## Summary

A Lighthouse audit of the production landing page (https://source.coop) found mobile performance at **54** while desktop sits at **94**. Desktop is genuinely excellent (LCP 0.8s, CLS 0). Every problem is a slow-network problem.

| Metric | Mobile (real 4G + 4× CPU) | Desktop |
|---|---|---|
| Performance | **54** | 94 |
| FCP | **6.1 s** (poor) | 0.4 s |
| LCP | **6.1 s** (poor) | 0.8 s |
| CLS | 0 (perfect) | 0 |
| TBT | 370 ms | 110 ms |
| Total bytes | **7,111 KiB** | |

## Root cause

First paint is gated by a single render-blocking stylesheet, and FCP tracks its arrival exactly:

| Scenario | Blocking CSS completes | FCP |
|---|---|---|
| As-is | 5901 ms | 6.1 s |
| Globe textures blocked | 5094 ms | 5.3 s |

`_next/static/css/1e5913857760d7aa.css` (77 KiB, `VeryHigh` priority) is **requested at 682 ms but does not arrive until 5.9 s**. It is starved on the shared HTTP/2 connection by everything requested alongside it:

- **6.5 MB of WebGL globe textures.** `clouds.png` takes **40.1 seconds** to download on 4G (698 ms → 40105 ms); `earth-blue-marble.jpg` takes 20.9 s. Both are `<link rel="preload">`'d in the marketing page head.
- **554 KiB of JavaScript across 31 chunks**, all requested in a 4 ms burst at ~728 ms.

Contention is severe enough that a 40 KiB self-hosted font at `High` priority takes 3.5 s to arrive. Berkeley Mono is not even *discovered* until 5247 ms, because it is declared inside the stylesheet that hasn't landed yet.

### Two hypotheses ruled out

- **Not CPU-bound.** Capping CPU slowdown at 1× left LCP unchanged (4.6 s).
- **Lantern hides this.** Under Lighthouse's default simulated throttling, blocking the images changed LCP not at all — simulation de-prioritizes `Low`-priority images out of the critical path. Only real (`--throttling-method=devtools`) throttling exposes the bandwidth contention. Use real throttling when re-measuring.

## Work items

- [ ] Remove the two `<link rel="preload">` globe texture hints — measured **800 ms** off both FCP and LCP — #472
[...]

PR #472, opened 2026-08-03
perf: drop globe texture preload hints from landing page

alukach:
Part of #471.

## What

Deletes the two `<link rel="preload">` hints for the WebGL globe textures in `src/app/(marketing)/page.tsx`.

## Why

The hints pulled **6.5 MB** of textures into the critical window, ahead of the render-blocking stylesheet that gates first paint. On a real-throttled 4G profile, `_next/static/css/1e5913857760d7aa.css` was requested at 682 ms but did not arrive until **5901 ms** — and FCP tracks that arrival exactly.

The preloads had no upside to weigh against that cost. The original comment said they warmed the cache "before the client component mounts", but `LiveGlobe` is `dynamic(..., { ssr: false })` — it mounts after hydration and requests the textures itself either way. The preload only changed *when* the bytes competed for bandwidth, not whether they were fetched.

## Measured impact

Lighthouse 13.4.1, mobile, `--throttling-method=devtools`:

| Metric | Before | After |
|---|---|---|
| FCP | 6.1 s | **5.3 s** |
| LCP | 6.1 s | **5.3 s** |
| Blocking CSS arrives | 5901 ms | 5094 ms |

Note that Lighthouse's *default* simulated throttling shows no change here — Lantern de-prioritizes `Low`-priority images out of the critical path and hides the contention. Re-measure with `--throttling-method=devtools`.

This is the first of several fixes; the textures are still oversized at 4096×2048, which #471 tracks separately.

## Risk

Low. Deletes two JSX lines, no behavior change — the globe loads its own textures on mount as before.

## Verification

- `npm run type-check` — 14 errors, identical to the count on clean `main` (all pre-existing, in `analytics/` files)

### DIFF
--- a/src/app/(marketing)/page.tsx
+++ b/src/app/(marketing)/page.tsx
@@ -43,9 +43,6 @@ export default async function Landing() {
 
   return (
     <>
-      {/* ponytail: no texture preload — LiveGlobe is dynamic({ ssr: false }), so it
-          mounts after hydration anyway. Preloading pulled 6.5MB into the critical
-          window ahead of the render-blocking CSS that gates first paint. */}
       <Box className={styles.landing} px={{ sm: "6", lg: "9" }}>
         <Box className={styles.landingInner} mx="auto" maxWidth="1400px">
           <Navigation />
--- a/src/app/(marketing)/page.tsx
+++ b/src/app/(marketing)/page.tsx
@@ -43,9 +43,9 @@ export default async function Landing() {
 
   return (
     <>
-      {/* Preload globe textures so they're cached before the client component mounts */}
-      <link rel="preload" href="/img/earth-blue-marble.jpg" as="image" />
-      <link rel="preload" href="/img/clouds.png" as="image" />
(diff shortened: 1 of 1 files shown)


src/app/(marketing)/page.tsx
```diff
@@ -43,9 +43,9 @@ export default async function Landing() {
 
   return (
     <>
-      {/* Preload globe textures so they're cached before the client component mounts */}
-      <link rel="preload" href="/img/earth-blue-marble.jpg" as="image" />
-      <link rel="preload" href="/img/clouds.png" as="image" />
+      {/* ponytail: no texture preload — LiveGlobe is dynamic({ ssr: false }), so it
+          mounts after hydration anyway. Preloading pulled 6.5MB into the critical
+          window ahead of the render-blocking CSS that gates first paint. */}
```
alukach:
```suggestion
```

jedsundwall:
Love refinements like this.

PR #475, opened 2026-08-03
perf: cache un-hashed static assets under /img and /logo

alukach:
Part of #471.

## What

Adds a `Cache-Control` rule for `/img/*` and `/logo/*` in `next.config.js`, and makes `headers()` apply in prod (it previously returned no rules at all there). The noindex rule stays non-prod only.

```
public, max-age=3600, stale-while-revalidate=604800
```

## Why

These assets are currently served `public, max-age=0, must-revalidate`, so every navigation pays a conditional-request round-trip for each one.

**Calibrating the win honestly:** I verified the revalidation returns `304` with **0 bytes**, so this was never a re-download problem — just latency. It's a modest fix, not a headline one, and it's the smallest item in #471.

## Why not `immutable`

The filenames under `public/` are not content-hashed, so `immutable` would be actively dangerous — a redeploy could never replace them for anyone holding a cached copy. That's not hypothetical: #473 changes the bytes of `clouds.png` at the same filename.

An hour of freshness removes the round-trip while bounding how long a replaced asset can linger, and `stale-while-revalidate` keeps the refresh off the critical path. Content-hashing these filenames would unlock a real `immutable` policy later; noted in #471.

## Verification

- `npx jest next.config.test.ts` — **4 passed**. I updated the three existing tests (the prod case asserted `[]`, which no longer holds) and added one asserting the cache rule applies in every stage.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -25,10 +25,26 @@ const nextConfig = {
   transpilePackages: ["jose"],
   serverExternalPackages: ["@duckdb/node-api"],
   async headers() {
+    // ponytail: un-hashed filenames under public/, so `immutable` is unsafe —
+    // a redeploy has to be able to replace these in place. An hour of freshness
+    // drops the per-navigation revalidation round-trip without stranding a
+    // stale asset for long.
+    const staticAssets = [
+      {
+        source: "/:dir(img|logo)/:path*",
+        headers: [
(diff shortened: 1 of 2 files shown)


alukach:
Not a priority

PR #476, opened 2026-08-03
fix: render zarr/icechunk preview in the full-width object preview card

alukach:
Addresses the review feedback on #415 ([review](https://github.com/source-cooperative/source.coop/pull/415#pullrequestreview-4846231557)). Stacked on `feat/zarr-icechunk-viewer` so it can be merged straight into that PR.

image

## The problem

#421 deliberately broke object previews out of the object list card so they can use the full width of the app container. This PR's Zarr/Icechunk viewer was rendering *inside* the Contents card, so it was confined to that card's width.

## What changed

- **Moved the store viewer into the `@preview` slot**, which already renders the full-width "Object Preview" card below the product grid. The main slot now just returns the directory listing for a store prefix (its files stay browsable, unchanged).
- **`StorePreview` → `storeViewerUrl()`.** The probe is now a helper returning the viewer URL or `null`, so the slot can decide whether to render the card at all. As a component it would have had to render *inside* an already-committed card, leaving an empty "Object Preview" card whenever a store failed the probe.
- **`PreviewIframe` is now shared** by file and store previews. Its duplicate "Open in new tab" link is gone — the card's `SectionHeader` already renders one, so the store preview would have shown two.

## Incidental fixes noticed along the way

- The slot's object path now strips a trailing slash. A trailing-slash URL (`/acct/prod/store.zarr/`) produces an empty final catch-all segment, which `getExtension` reads as "no extension" — so the viewer would silently not appear. The main slot already normalized this; the slot did not.
- Dropped `isViewableStorePath`, which was exported and tested but never called. Its trailing-slash coverage is folded into the `isStoreExtension` test.

## Verification

- `npx tsc --noEmit` — clean for every touched file.
- `npm run lint` — no new warnings.
- Jest: `page.test.tsx` (13), `storeViewer.test.ts`, `files.test.ts`, `probe.test.ts` all pass. The main-slot store test now asserts the listing renders alone, since the viewer moved out.
- `next build` compiles the route graph successfully. Note it then fails on pre-existing type errors in `src/components/features/analytics/` (byte-identical to `origin/main`, not from this change) and on prerendering `/` against local AWS creds.

**Not verified in a browser** — I could not render the preview against a live Zarr store locally, so the visual result is unconfirmed. Worth a look at the Vercel preview before merging.

### DIFF
--- a/src/app/(app)/[account_id]/[product_id]/(product)/[[...path]]/@preview/page.tsx
+++ b/src/app/(app)/[account_id]/[product_id]/(product)/[[...path]]/@preview/page.tsx
@@ -1,13 +1,16 @@
-import { Suspense } from "react";
-import { Card, Flex, Link } from "@radix-ui/themes";
+import { Suspense, type ReactNode } from "react";
+import { Box, Card, Flex, Link } from "@radix-ui/themes";
 import { SectionHeader } from "@/components/core/SectionHeader";
 import { getStorageClient } from "@/lib/clients/storage";
 import { readProxyCredentials } from "@/lib/services/proxy-credentials-read";
+import type { ProxyCredentials } from "@/lib/actions/proxy-credentials";
 import { ObjectPreview } from "@/components/features/products/object-browser/ObjectPreview";
+import { PreviewIframe } from "@/components/features/products/object-browser/PreviewIframe";
+import { storeViewerUrl } from "@/components/features/products/object-browser/storeViewer";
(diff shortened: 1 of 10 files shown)


alukach:
Superseded: the fix is now commit 58cb9fe4 directly on `feat/zarr-icechunk-viewer`, so it's part of #415. Closing this to collapse the stack — one PR to review instead of two, and no stack rebase needed.

PR #488, opened 2026-08-17
fix(products): validate product_id as a lowercase slug

alukach:
`product_id` was the only ID in the codebase declared as a bare `z.string()` (`src/types/product.ts:56`). Every other one — account, membership, API key, data connection — runs through `MIN/MAX_ID_LENGTH` + `.toLowerCase()` + `ID_REGEX` from `src/types/shared.ts`. Products alone accepted capitals, spaces, underscores, slashes, and any length.

The product form's client-side check (`useProductIdValidation`) disables the submit button for non-slug IDs, but `createProduct` is a server action — an HTTP endpoint a caller can post to directly — and validated only with `ProductCreationRequestSchema`. The v1 API `POST /products/{account_id}` route is stubbed `501`, so the server action was the whole surface. DynamoDB keys are case-sensitive, so `alice/MyProduct` was only ever reachable at that exact spelling.

## Change

`product_id` now uses the same rules as `account_id`, via a new `ProductIdSchema`. It flows through `ProductCreationRequestSchema` into the create path, and the mirror prefix is derived from the normalized value.

## Tests

Eight new cases in `src/lib/actions/products.test.ts`, exercising the server action rather than the schema, since the action is the boundary that was bypassable. All eight fail on `main` and pass here:

- `MyProduct` → stored as `myproduct`, mirror prefix `alice/myproduct/`
- rejects space, underscore, slash, leading hyphen, consecutive hyphens, <3 chars, >40 chars

## Existing data

Legacy rows with non-conforming IDs are unaffected: reads cast to `Product` rather than parse, and `updateProduct` reads `product_id` from the form without a schema, so such products stay editable and viewable. Worth a table scan to confirm none exist before assuming the invariant holds going forward.

## Verification

`npx jest src/lib/actions/products.test.ts` — 31 passed. Full suite: 489 passed, 5 pre-existing failures in the analytics/recharts and DropdownSection suites (identical with this change stashed). `tsc --noEmit` reports no errors in the changed files (pre-existing errors elsewhere in analytics).

### DIFF
--- a/src/lib/actions/products.test.ts
+++ b/src/lib/actions/products.test.ts
@@ -151,6 +151,47 @@ describe("createProduct", () => {
     expect(productsTable.create).not.toHaveBeenCalled();
   });
 
+  // The product form blocks non-slug ids client-side, but a server action is
+  // just an HTTP endpoint — these cover a caller posting to it directly.
+  test("lowercases a capitalized product ID", async () => {
+    (dataConnectionsTable.fetchById as jest.Mock).mockResolvedValue(
+      connection({
+        prefix_template:
+          "{{repository.account_id}}/{{repository.repository_id}}/",
+      })
+    );
+
+    const result = await createProduct(
+      undefined,
+      buildFormData({ product_id: "MyProduct" })
+    );
+
+    expect(result.success).toBe(true);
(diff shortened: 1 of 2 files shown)


alukach:
closed in favor of #489

Issue #491, opened 2026-08-20
Epic: machine identity for unattended workflows


alukach:
## The problem

Source Cooperative can't authenticate software acting on a user's behalf. Getting credentials requires a human at a browser. So anything unattended — a nightly sync, a publishing pipeline, an instrument uploading observations — has two bad options: babysit a login, or embed a person's session somewhere it doesn't belong.

This epic adds **service accounts**: a login for software, that a user creates, grants, and revokes without borrowing anyone's account.

**Scope is deliberately narrow — reading and writing product data only.** A service account does not manage application data: no creating products or organizations, no managing members, and it can never own a record.

---

## What a user actually does

This is the whole feature from the outside.

**1. Create a service account.** From your user or organization settings: give it a name. It belongs to you, or to your organization.

**2. Say how software proves it's this account.** You can add more than one, and add or remove them later:

- **A GitHub repository** — "the `main` branch of `myorg/myrepo`." No secret to store anywhere; GitHub vouches for the workflow.
- **An API key** — a secret Source issues once, for anything that isn't on a platform that can vouch for itself. Shown once, revocable, always has an expiry date.

Both routes lead to the same account with the same access. If you need CI to have *different* access than a key, make two service accounts.

**3. Say what it can reach.** Either everything under your account, or specific products. For each, read or write.

**4. Tick which roles it may use.** When a workflow requests credentials for a service account, it must also request a role to use during that session. A role can only ever take access away, never add it — so this is how you say "this account may never write, whatever its grants say," and it's also how a single account can serve a deploy job and a read-only analysis job. **Full access** and **Read only**, both ticked by default. A workflow utilizing the **Read only** role can't edit data, even if the service account is permitted to do so.

**5. Point your software at it.** For GitHub Actions, three lines of workflow config and no secrets. For anything else, install the key and let the Source CLI keep it fresh. Your software names the role it wants when it asks for credentials.

That's it. Revoke a sign-in method and that route stops working. Revoke a grant and access stops within a minute. Disable the service account and everything stops.

---

## How it works

Four things, each with one job:

| | What it answers | Example |
|---|---|---|
| **The service account** | *Who is this?* | `svc--nightly-sync`, owned by `noaa` |
| **Sign-in methods** (one or many) | *How does it prove that?* | this GitHub repo; this API key |
| **Grants** (one or many) | *What may it touch?* | write on `noaa/buoys`, read on `noaa/tides` |
[...]

PR #495, opened 2026-08-22
feat(settings): mock service account creation form for #491

alukach:
## What this is

A **non-persisting design mock** of the service-account creation flow proposed in #491, so the data model can be reviewed in a browser before we build it.

Nothing is saved. Submitting the form renders the database rows the creation *would* write.

## Where to find it

**Settings → Service Accounts**, on both individual and organization accounts (either can own a service account). Visible to anyone who can manage the account.

## What it demonstrates

The four decisions from the epic, in order:

1. **Name** — a human-friendly name; the stored id is derived from it and previewed live as `svc--<sanitized>`. Sanitization collapses double hyphens, so a name can never forge a second reserved prefix (`ID_REGEX` forbids `--` in human account ids, which is what makes the prefix collision-proof).
2. **How software signs in** — one or many. A GitHub repository + ref (no secret stored), or an API key that expires in 30/90/365 days *or never*. Every route reaches the same account with the same access.
3. **What it can reach** — the whole owner account, or specific products, each read or read-write.
4. **Which roles it may use** — Full access and Read only, both ticked by default. The form notes that both-ticked is a no-op and that unticking Full access is the durable restriction.

Submitting lands on a **mock detail view** of the integration — the page a real create would redirect to — with an **Edit** button that returns to the form with every answer intact. If a key sign-in method was chosen, the detail view shows the **one-time API key** with copy-now framing.

The detail view shows:

- **`accounts`** *(existing table)* — the new `service` row with `owner_account_id` and `allowed_roles`
- **`identity_bindings`** *(new table)* — one row per sign-in method, all pointing at the same principal
- **`memberships`** *(existing table)* — one row per grant; account-wide access collapses to a single row with no `repository_id`, which is already how `hasRole` reads it
- Ready-to-paste config per sign-in method: a GitHub Actions workflow using **`aws-actions/configure-aws-credentials@v4`**, or the CLI + env-var form for keys. The role in both changes if Full access is unticked.

Plus the caveats worth seeing next to the model: what doesn't exist yet, that roles only subtract, and that revoking a grant stops new access in ~60s while issued credentials last up to an hour.

## Notes for review

- `plan.ts` is pure and holds all the model logic; `plan.test.ts` covers it (14 tests) — namespacing, the GitHub `sub` claim shape, per-product vs account-wide membership rows, one binding per sign-in method, and the validation rules including the empty-role-set case that would leave an account unable to authenticate at all.
- The page is gated on `PutAccountProfile`, not `ListAccountMemberships` — the latter returns `true` for any signed-in user (`authz.ts:1242-1252`), which would have exposed the mock on other people's accounts.
- No API routes, no server actions, no schema changes. The only edits to existing files are a nav entry and a URL helper.

## Two things the mock deliberately surfaces for decision

- **`aws-actions/configure-aws-credentials` also calls `GetCallerIdentity`**, which `/.sts` does not implement. Since that action is how most people will first try this, the mock shows it *and* flags the gap.
- **Keys with no expiry** are offered, but #491 currently says keys cannot be issued without one. The mock calls this out as a departure rather than quietly picking a side.

## Not included

Key issuance, revocation, the `delete` permission question, and custom roles. This is the create flow only.

https://claude.ai/code/session_01HGZ598AJ1F7NWoyPpgepBt


alukach:
Some design concepts for the cards:

Main@1x
StatusRail@1x
SpecSheet@1x

Going with option C for now.

PR #497, opened 2026-08-24
refactor(forms): one Field primitive across every form

alukach:
Performs the refactor from the [design canvas](https://claude.ai/code/artifact/66c8a032-8f0d-474f-86d6-09a5a9af24f6).

**Storybook has been split out into #498, stacked on this branch.** This PR is now forms only. (The branch was rewritten to drop the Storybook commits, so the two earlier review threads no longer anchor to lines — both findings were addressed, and the responses are in the comments below.)

## Why

The field anatomy — label, help, control, error — had already been written twice: inlined across three branches of `DynamicForm`, and again as a local `Field` in `DataConnectionForm`, which imported `formFieldStyle` across a module boundary to keep the two looking alike. So this is mostly deletion.

`formFieldStyle` hand-rolled what the Radix theme already provides. Removing it is the whole of "use the existing Radix theme": focus rings, disabled states, dark mode and the configured `radius: none` / gray accent / 110% scaling now apply because nothing overrides them.

## What changed

**`<Field>`** (`src/components/core/Field.tsx`) owns the wiring nobody remembers: generated id, label association, `aria-describedby` covering help *and* errors, `aria-invalid`. The control is a **child**, not a `type` union — which is what lets one wrapper serve a text input, a select, radio cards, a switch or a dropzone.

Where there is no single labelable control, the field is marked `group` and the label names a `role="group"` via `aria-labelledby`. This matters more than it sounds: `htmlFor` only associates with input, select, textarea, button, meter, output and progress. Pointing it at `RadioCards.Root` (a `div[role=radiogroup]`) resolves to a node that names nothing, and a `customComponent` can never receive the id at all.

**`<FormActions>`** gives a form one action row. `InviteMemberForm` rendered Cancel as a sibling of the form, so the dialog had two right-aligned rows with Cancel stranded under the primary button.

**`DynamicForm`** keeps its public API — the seven forms using it needed no changes — but renders Radix controls through `Field` and groups fields by `section`. Two type-level guards were added: `radio-cards` and `switch` render entirely from their value prop, so the field type now requires `controlled`/`value`/`onValueChange` rather than letting someone ship a control that never moves when clicked. The character counter tracks live input inside `DynamicForm` instead of reading the seeded value, so it doesn't silently depend on the caller also wiring the field as controlled.

**Per-form**, from the canvas: product visibility is radio cards (a disabled `<option>` can't say *why* it's unavailable; a card can), status is a Switch instead of a two-value select, and delete moves into a labelled `DangerZone`.

## One bug fixed on the way

The bio help text promised "220 characters maximum". `AccountProfileSchema` enforces 1024. Nobody wrote a bug — two sources of truth drifted and nothing could catch it. `BIO_MAX_LENGTH` is now exported from the schema and drives the counter, so they can't drift again.

## Guardrail

An eslint `no-restricted-syntax` rule keeps raw `input`/`select`/`textarea` out of `src/components`, exempting `type="hidden"` and `type="file"`. One inline disable, in `Dropzone`, where `react-dropzone` supplies the type through a spread.

## Tests

`Field` in isolation, plus `DynamicForm`'s field-type dispatch — including a general check that no `label[for]` in a rendered form points at a non-labelable element, which encodes the rule rather than the instance.

Worth knowing why that dispatch was untested before: `DynamicForm` could not be rendered under jest at all. It calls `useActionState`, which doesn't exist in the `react@18.3.1` jest resolves — the app only gets that hook through Next's bundled runtime. That gap is why two rounds of label-wiring bugs reached review. A shim in the test file adds the missing export to the real module (neither a spread nor a Proxy survives babel's interop, which enumerates own keys).

## Verification

`tsc --noEmit` clean · jest 59 suites / 505 tests · `next lint` 0 errors.

`next build` compiles and lints clean but can't finish locally: prerendering the marketing homepage calls DynamoDB and there are no AWS credentials in this environment. That path is not in this diff.

## Follow-up, not here

The primary button fails WCAG AA contrast — white on `--accent-9` (`#8d8d8d`, 3.31:1 against a required 4.5:1) because the theme sets `accentColor="gray"`. That's every primary button in the app, not something this PR introduced, and it wants its own change.

https://claude.ai/code/session_011mnsg8m1dXZ5z5ig5xMtkE

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,13 +5,18 @@
     "plugin:@typescript-eslint/recommended",
     "next/core-web-vitals"
   ],
-  "plugins": ["@typescript-eslint"],
+  "plugins": [
+    "@typescript-eslint"
+  ],
   "rules": {
     "@typescript-eslint/no-explicit-any": "warn",
-    "@typescript-eslint/no-unused-vars": ["warn", { 
-      "argsIgnorePattern": "^_",
-      "varsIgnorePattern": "^_"
-    }],
+    "@typescript-eslint/no-unused-vars": [
+      "warn",
+      {
+        "argsIgnorePattern": "^_",
+        "varsIgnorePattern": "^_"
+      }
+    ],
     "@typescript-eslint/no-require-imports": "warn",
     "prefer-const": [
       "error",
@@ -20,5 +25,33 @@
       }
     ]
   },
-  "ignorePatterns": ["node_modules/", ".next/", "public/", "**/*.d.ts"]
+  "overrides": [
+    {
+      "files": [
+        "src/components/**/*.tsx"
+      ],
+      "excludedFiles": [
+        "**/__tests__/**"
+      ],
+      "rules": {
+        "no-restricted-syntax": [
+          "error",
+          {
+            "selector": "JSXOpeningElement[name.name=/^(select|textarea)$/]",
+            "message": "Use Radix Themes Select/TextArea inside a <Field>. A raw control skips the theme (focus ring, disabled state, dark mode)."
+          },
+          {
+            "selector": "JSXOpeningElement[name.name=\"input\"]:not(:has(JSXAttribute[name.name=\"type\"][value.value=/^(hidden|file)$/]))",
+            "message": "Use Radix Themes TextField/Checkbox/Switch inside a <Field>. A raw control skips the theme (focus ring, disabled state, dark mode). type=\"hidden\" and type=\"file\" are exempt."
+          }
+        ]
+      }
+    }
+  ],
+  "ignorePatterns": [
+    "node_modules/",
+    ".next/",
+    "public/",
+    "**/*.d.ts"
+  ]
 }
--- a/src/app/(app)/edit/product/[account_id]/[product_id]/details/page.tsx
+++ b/src/app/(app)/edit/product/[account_id]/[product_id]/details/page.tsx
@@ -14,7 +14,7 @@ import {
   DataConnection,
   DataConnectionObjectSchema,
 } from "@/types";
-import { Box, Separator, Text } from "@radix-ui/themes";
+import { DangerZone } from "@/components/core";
 
 export async function generateMetadata({
   params,
@@ -75,21 +75,18 @@ export default async function DetailsPage({ params }: PageProps) {
       />
 
       {canDelete && (
-        <>
-          <Separator size="4" my="6" />
-          <Box>
-            <Text as="p" size="2" color="gray" mb="3">
-              Deleting this product is permanent and cannot be undone. All
-              associated data, memberships, and records will be removed.
-            </Text>
+        <DangerZone
+          title="Delete this product"
+          description="Permanent. Removes the product record, its memberships and its access records. Whether the underlying objects go too depends on the data connection."
+          action={
             <DeleteProductModal
               accountId={account_id}
               productId={product_id}
               canPreserveData={canPreserveData}
               dataReadOnly={!!connection?.read_only}
             />
-          </Box>
-        </>
+          }
+        />
       )}
     </>
   );
--- a/src/components/core/DangerZone.tsx
+++ b/src/components/core/DangerZone.tsx
@@ -1,0 +1,42 @@
+import { Box, Flex, Text } from "@radix-ui/themes";
+
+interface DangerZoneProps {
+  title: string;
+  description: React.ReactNode;
+  /** The destructive control itself. */
+  action: React.ReactNode;
+}
+
+/**
+ * Where irreversible actions live. Bordered and tinted so a delete control is
+ * never mistaken for the rest of the form.
+ */
+export function DangerZone({ title, description, action }: DangerZoneProps) {
+  return (
+    <Box mt="6">
+      <Text as="p" size="2" weight="bold" color="red" mb="2">
+        Danger zone
+      </Text>
+      <Flex
+        align="start"
+        justify="between"
+        gap="5"
+        p="4"
+        style={{
+          border: "1px solid var(--red-6)",
+          backgroundColor: "var(--red-2)",
+        }}
+      >
+        <Box>
+          <Text as="p" size="2" weight="medium">
+            {title}
+          </Text>
+          <Text as="p" size="1" color="gray" mt="1">
+            {description}
+          </Text>
+        </Box>
+        <Box flexShrink="0">{action}</Box>
+      </Flex>
+    </Box>
+  );
+}
--- a/src/components/core/DynamicForm.tsx
+++ b/src/components/core/DynamicForm.tsx
@@ -1,33 +1,97 @@
 "use client";
 
 import React, { useActionState, startTransition } from "react";
-import { Button, Text, Flex } from "@radix-ui/themes";
+import {
+  Flex,
+  RadioCards,
+  Select,
+  Switch,
+  Text,
+  TextArea,
+  TextField,
+} from "@radix-ui/themes";
 import { useRouter } from "next/navigation";
+import { Field } from "./Field";
+import { FormActions } from "./FormActions";
+import { SectionHeader } from "./SectionHeader";
 
-export interface FormField<T extends Record<string, any>> {
+/** Controls that render purely from a value prop, so they must be controlled. */
+export type ValueDrivenFieldType = "radio-cards" | "switch";
+
+export type StandardFieldType =
+  | "text"
+  | "textarea"
+  | "email"
+  | "url"
+  | "password"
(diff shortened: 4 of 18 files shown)


alukach:
Both findings were real. Verified in the code before acting on them.

**1. `DataConnectionForm` never wired `Field` — confirmed, and worse than described.**

The reviewer is right about the cause: my transform swapped the controls to Radix but left children as plain elements, and `Field` only spreads the wiring through its render-prop form. Worth noting it was a regression, not just an omission — the old local `Field` rendered a plain `<Text>`, so nothing claimed an association. The new one renders `<label htmlFor={id}>` against an id that exists on nothing, which is a *broken* association rather than an absent one.

Fixed: 24 single-control fields now use the render prop.

I diverged on the three the reviewer set aside. The two checkbox groups are fine *internally* — each checkbox has its own native `<label>` wrapper — but the enclosing `Field` still rendered "Read Only" / "Allowed Visibilities" as a `<label htmlFor>` pointing nowhere, so the same defect remained one level up. Same for "Trust-policy subject", which is a read-only `<Code>` display. Rather than point `htmlFor` at an arbitrary member of the set, those three now use a new `group` prop: the label becomes the accessible name of a `role="group"` via `aria-labelledby`, which is what a set of controls actually needs.

The reviewer's sharpest point was that `Field.stories.tsx` modelled the unwired pattern in its own primary examples, so it read as sanctioned usage. Those are rewired too — single controls to the render prop, sets to `group`.

Added a test for the `group` path specifically, since that's the shape that regressed silently. `Field` is now 7 tests.

**2. `radio-cards`/`switch` no-op when uncontrolled — confirmed, took the type-level option.**

`FormField` is now a discriminated union: those two types require `controlled: true`, `value` and `onValueChange`. Unrepresentable rather than discoverable. It immediately caught two of my own stories, which is the best evidence it was worth doing.

**Also fixed, from running the Storybook I'd shipped** (these landed in the same commit as the wiring fix, `71d4bf0b`, and its message doesn't mention them — noting it here rather than rewriting pushed history):

- `DynamicForm` calls `useRouter`, and without `parameters.nextjs.appDirectory` the app router mock is never mounted — every `DynamicForm` story threw `invariant expected app router to be mounted`.
- `globals.css` points `--default-font-family` at `--font-ibm-plex`, which only exists because the root layout puts next/font's generated class on `<body>`. Storybook had no such class, so that `var()` was invalid at computed-value time and every control fell back to the browser serif. Preview now loads the same face and applies the same class.

Neither was caught by `storybook build`, which compiles stories without rendering them.

**Verification:** `tsc --noEmit` clean · jest 58 suites / 495 tests · `next lint` 0 errors · `storybook build` succeeds.

alukach:
Both correct, and both the same bug class as the first review — one level up, in `DynamicForm`'s dispatch rather than in a form calling `Field`.

**`radio-cards`** — confirmed. `RadioCards.Root` renders `div[role=radiogroup]`, and `htmlFor` only associates with *labelable* elements (input, select, textarea, button, meter, output, progress). The id resolved to a node, so it didn't look dangling, but it named nothing. `ProductCreationForm`'s "Visibility" had no accessible name.

**`type: "custom"`** — confirmed. `renderControl` returns `customComponent` as-is and can't hand it the id at all, so `EditProfileForm`'s "Websites" pointed at nothing. A `ReactNode` can never receive the props, which means `htmlFor` could *never* be right for this type — so it's unconditionally a group case, as you say.

Both now route through `Field`'s `group`, and `Field`'s render-prop branch passes `aria-labelledby` as you sketched. A labelled custom field is additionally wrapped in `role="group"` to carry the naming, since the component itself can't take the attribute.

I checked `switch` while I was in there — that one is genuinely fine. Radix renders `<button role="switch">`, and `button` is labelable.

**On the coverage gap, which is the more useful half of this review.** The reason both escaped is worse than "Field's tests only exercise Field": `DynamicForm` could not be rendered under jest *at all*. It calls `useActionState`, which doesn't exist in the `react@18.3.1` jest resolves — the app only gets it through Next's bundled runtime. So the dispatch that decides which control each field type gets had zero coverage, and that's exactly where both of these bugs lived.

Fixed by adding the missing export to the real react module inside a mock. Worth recording that neither a spread nor a Proxy works — babel's interop enumerates own keys, so a virtual property never survives it, and a copy loses `default`/`__esModule`.

That buys 7 dispatch tests covering every field type, including a general one asserting that no `label[for]` in a rendered form points at a non-labelable element — which encodes the rule rather than the instance. I verified all three relevant tests fail with the fix reverted; a test that passes either way would have been worthless here.

Not changed: `a11y test: "todo"` stays as-is. Flipping it to `"error"` would fail CI today on a genuine pre-existing violation — the primary button is white on `--accent-9` (`#8d8d8d`, 3.31:1 against a required 4.5:1) because the theme sets `accentColor="gray"`. That's every primary button in the app, not something this PR introduced, and it wants its own change.

**Verification:** `tsc` clean · jest 59 suites / 504 tests · `next lint` 0 errors · `storybook build` succeeds.

alukach:
Thanks — acting on the non-blocking finding, and taking it a step further than suggested.

**The gap is real.** `AccountFlagsForm` is exactly the shape described: each capability sets `description` with no `label`, so every flag's explanation rendered with an id nothing referenced.

**I didn't take the suggested fix as-is, though.** Widening the wrapper condition to `field.label || aria-describedby || aria-invalid` stops the props being *dropped*, but it doesn't make the description reach the control: `aria-describedby` on a wrapper is not announced with the checkbox inside it. Screen readers announce a control's own description. So the group wrapper was always the weaker half of the fix.

`customComponent` now also accepts a **function**, receiving the same ids `Field` hands its render-prop children — the same pattern that fixed `DataConnectionForm` earlier in this PR. `AccountFlagsForm` uses it, so each `Checkbox` carries `aria-describedby` itself and the flag description is genuinely associated.

The wrapper stays as the fallback for a plain `ReactNode`, with exactly the widened condition you suggested, so nothing is silently dropped for callers that can't take the props. Both paths are tested, and I verified both fail with the change reverted — the described-but-unlabelled case and the function form.

Unrelated to the review, from testing against the running app in the meantime:

- Radio cards had no selection indicator at all — Radix's `RadioCards` conveys selection by card border alone. Added one drawn from Radix's own `Radio` tokens (`--accent-indicator`, `--accent-contrast`, `--gray-a3/a8`), decorative and `aria-hidden` so `RadioCards.Item` stays the only radio.
- `.rt-RadioCardsItem` sets `align-items: center` *and* `justify-content: center`, so with uneven descriptions the shortest card floated mid-height and indented from the left. Both axes now anchor to the start.
- A disabled option no longer restates why it's unavailable. The greyed treatment carries it and the field help names the dependency once; repeating `Not offered by <connection>` on every blocked card was noise, badly so with long connection names.
- `allowedVisibilitiesFor` used `connection?.allowed_visibilities?.length` as its fallback test, conflating "couldn't resolve the connection" with "connection permits nothing" — so an empty `allowed_visibilities` offered all three options while the server rejected every one. Fixed, and such a connection is excluded from the create picker.
- The connection form's "(Not currently enforced.)" copy on `allowed_visibilities` and `required_flag` was stale. Both **are** enforced — in `createProduct`/`updateProduct` and `useDataConnection` respectively. The only place the checks are absent is the v1 API product POST, which returns 501 and is dead code.

**Verification:** `tsc` clean · jest 59 suites / 507 tests · `next lint` 0 errors.

alukach:
Both correct — and both were dead code I'd left behind, so I deleted them rather than annotating them.

**`visibilityOptions`.** The guard made sense when the base list was the connection's `allowedVisibilities`: the current value could genuinely fall outside it. Once the list became `ALL_VISIBILITIES` the condition could never fire, and worse, its comment still claimed to keep an out-of-policy current value selectable — protection that had already moved to the `disabled` predicate (`!permitted && value !== visibility`). So it was dead *and* misleading about where the real behaviour lives. Dropped the guard, moved the explanation to the predicate that actually implements it.

**The uncontrolled fallbacks** in the `radio-cards` and `switch` branches. `ValueDrivenFormField` requires `controlled: true` and a `value`, so the discriminant narrows both branches and the fallback can't run. `tsc` accepting a bare `field.value` with no fallback is the proof — which is a nice property of having made that constraint type-level in the first place: the compiler can now confirm the dead branch is dead.

I'd push back gently on "fine as defensive code" for the second one. A fallback that can't execute isn't defence, it's a claim that the invariant might not hold — which invites the next reader to weaken the type to match. Deleting it keeps the type as the single source of truth.

**Verification:** `tsc` clean · jest 59 suites / 507 tests · `next lint` 0 errors · `storybook build` succeeds on the stacked #498.

PR #508, opened 2026-08-25
fix(authz): allow org owners to create products without CREATE_REPOSITORIES flag

tylere:
## What I'm changing

Org-wide owners and maintainers were blocked from creating products under their organization because `createRepository()` checked the `CREATE_REPOSITORIES` account flag before checking the user's org role. This meant the `/products/new` page redirected them away even though the underlying server action would have succeeded.

## How I did it

Added `hasRole()` shortcuts in `createRepository()` before the flag check: one for the specific-product path (when the product belongs to a different account) and one for the wildcard path (`"*"`, used by the `/products/new` page gate) that scans the user's memberships for any active org-wide owner or maintainer role. The flag check is now a fallback only for users who are neither org owners nor own the target account.

## How you can test it

Log in as an org owner whose personal account does not have the `CREATE_REPOSITORIES` flag. Confirm you can navigate to `/products/new` without being redirected and successfully create a product under your organization. The new unit test in `authz.test.ts` covers both the specific-product and wildcard paths.


alukach:
@claude extend this PR to also only show "New Product" button on account list dropdown and to filter the dropdown of accounts in the product creation form to only those where the user has creation permissions

src/lib/api/authz.ts
```diff
@@ -1095,7 +1115,37 @@ function createRepository(
...
+      (m) =>
+        m.state === MembershipState.Member &&
+        !m.repository_id &&
+        (m.role === MembershipRole.Owners ||
+          m.role === MembershipRole.Maintainers)
+    );
+    if (hasOrgRole) {
+      return true;
+    }
+  }
```
alukach, line 1146: @claude Verify that this is right. It seems like it's looking for being an owner/maintainer of _any_ org. Not sure when this would be useful in the application

alukach (approved):
This PR represents a departure in that it now allows any mantainer/owner of an org membership to create products, ie org admin status implies the right to create a product rather than us granting someone that right explicitly. I think that's fine, if we are creating an org we almost surely want products within it, but I think it's worth calling that change out directly.

PR #511, opened 2026-08-25
fix(storage): fall back to range GET when proxy strips Content-Length

### LINKED ISSUE #509: Object sizes reported as '0 B' for some files
## Description

Some objects are displaying incorrect sizes in the file browser. For example, [geospatialml/fmow/index.html](https://source.coop/geospatialml/fmow/index.html) shows a size of **0 B**, but downloading the file reveals it is approximately 5 KB.

## Steps to Reproduce

1. Navigate to https://source.coop/geospatialml/fmow/index.html
2. Observe the file size column — it shows **0 B**
3. Download the file and check its actual size — it is ~5 KB

## Expected Behavior

The file size column should display the actual size of the object (e.g., ~5 KB).

## Actual Behavior

The file size is shown as **0 B**.

## Screenshot

Object size shown as 0 B

## Notes

This may be related to how object metadata (e.g., `Content-Length`) is retrieved or parsed. Objects with a size of 0 in the metadata but a non-zero actual size could indicate the size is not being fetched or stored correctly, or that the wrong metadata field is being used.

tylere:
## Summary

- The data proxy omits `Content-Length` from HEAD responses, so `getObjectInfo()` always sees `ContentLength` as `undefined` → `0`, and `ObjectSummary` displays **0 B** for every object
- When `ContentLength` is absent (or `0`) from the HEAD, issue a `GET` with `Range: bytes=0-0` and read the actual total from the `Content-Range` response header (`bytes 0-0/TOTAL`)
- Failures in the fallback are swallowed so objects on proxies that don't support range requests still return gracefully with `size: 0`

Fixes #509

## Test plan

- [x] Existing `getObjectInfo` tests still pass
- [x] New test: falls back to range GET when HEAD omits `Content-Length`
- [x] New test: falls back to range GET when `ContentLength` is `0`
- [x] New test: keeps `size` as `0` when the range GET also fails
- [ ] Manual: visit https://source.coop/geospatialml/fmow/index.html after deploy — Size should show ~5 KB instead of 0 B


alukach:
@claude address the PR review (https://github.com/source-cooperative/source.coop/pull/511#issuecomment-5415812224) and also consider if we could avoid interacting with the request body in entirety and instead only review the response headers

PR #557, opened 2026-09-15
feat(analytics): scope the product usage card to the current prefix

### LINKED ISSUE #556: [Proposed Feature] Filter analytics by prefix as user navigates through product
**Description of Feature:**

When viewing a product, the analytics stay visible despite the fact that the may navigate into a prefix within the product.

| Image |
| --- |
| Analytics for entire product visible on root of product |

Image |
| --- |
| Analytics for entire product visible on `docs/` prefix of product |

Instead, the analytics card should update to demonstrate:

1. analytics for all requests beginning with that prefix
2. the prefix that the analytics relate to

### Costs

Currently, we cache analytics fetches from Cloudflare Workers Analytics Engine to minimize costs. This would create more values that should be cached and more requests made to the Analytics Engine. However, I don't believe this will be a substantial cost center as lookups are currently priced @ $1/million reads after the first million (https://developers.cloudflare.com/analytics/analytics-engine/pricing/)

More insights to a user for users who seek sub-product aggregated analytics information.

alukach:
Closes #556.

The analytics card stayed on the product's whole-product numbers however deep into the product the viewer navigated, so a prefix's traffic was unreadable and the card on `docs/` was indistinguishable from the one on the product root.

The card now follows the viewer: at `/{account}/{product}/docs/2026` the numbers cover `docs/2026` and everything under it, and the path is named above them, slash-terminated so it reads as a prefix rather than as one object.

## How

`getUsage`'s third argument was an exact-match object key (`blob3 = '…'`) with no production caller — only tests. It is now a path filter, `(blob3 = 'docs' OR blob3 LIKE 'docs/%')`, which is one object or a whole directory without the caller having to know which the path is. That matters because the card stays mounted on object pages too, where the "prefix" is a file. A bare `LIKE 'docs%'` would have been shorter and wrong: it counts downloads of the sibling `docsets.tif`.

The product layout passes the decoded path — the data proxy decodes the key before logging it (`RequestParts` decodes for bucket/key routing), so `blob3` holds decoded keys, the same coordinates the directory listing uses.

The manager-only `/-/analytics` page is unchanged in what it reports: the whole product.

One thing to weigh in review: the label always ends in a slash, including on an object page, where the path is a file. Telling a file from a directory in the layout would cost a storage lookup the layout does not otherwise make, and the filter is right either way — but `data.tif/` is a slightly odd label for an object.

## The card's query budget

The first version of this branch made the card disappear on prefix pages, with `Analytics Engine query failed (429)` in the logs. Worth recording, because the cause is a property of the feature rather than a typo: scoping to a prefix makes **every directory its own cache key**, and each cold one was six Analytics Engine queries — the day series, the window totals, and four weekly per-IP slices. Browsing a product turns into a burst of uncached SQL API calls, and the API starts refusing them.

On the limit itself: [Analytics Engine's limits page](https://developers.cloudflare.com/analytics/analytics-engine/limits/) documents writes and retention only — no query rate limit, and no mention of 429. These queries go to `api.cloudflare.com`, whose [general API limits](https://developers.cloudflare.com/fundamentals/api/reference/limits/) are 1,200 requests per five minutes per user/token, after which "all API calls for the next five minutes will be blocked". That would explain the observed behaviour, including the card vanishing product-wide rather than only where the budget was spent, but Cloudflare does not document whether SQL API calls draw on that bucket, so treat it as the likely explanation rather than an established one. The 429 is the fact; the budget it came from is inference. A 429 now logs its `Retry-After`, so the next occurrence says more than this one could.

Four of those six queries fed `users` (unique IPs, the signed-in split, the frequency histogram), which nothing on the product page renders — only the manager-only analytics page does. So they move to `getUsageUsers`, which that page calls alongside `getUsage`:

- product card, per prefix: **2 queries** (was 6)
- analytics page, per product: 7 for a 28-day window (was 6), once per product per revalidate

`ProductAnalyticsView.users` becomes nullable as a result — the two fetches fail independently, and the USERS tab now says so in a line instead of taking the whole page down with it.

A 429 also parks the client for `Retry-After` (a minute by default) before it tries again. Failures aren't cached — `unstable_cache` stores results, not throws — so without that, every render keeps hitting an API that is already refusing, which at best wastes the recovery window and at worst holds a limiter open. It is per server instance, not fleet-wide; that is marked with a `ponytail:` comment rather than reached for a shared store.

Reads remain $1/million after the first million, as the issue anticipated; the constraint that bit is request rate, not price.

Two things ruled out along the way, so they don't get re-investigated:

**Link prefetching is not the multiplier.** A directory listing makes Next.js issue one RSC prefetch per entry — 26 for one `window_b` listing, 20 of them individual `.tif` files. Replaying those exact requests (`RSC: 1` + `Next-Router-Prefetch: 1`) and diffing against the same path without the prefetch header shows the prefetch returns 57 KB containing neither `Analytics`, `Contents` nor `product-meta`, against 113 KB containing all three on a real navigation. `loading.tsx` sits above this layout, so a prefetch stops at that shell and `UsageCard` never runs. Those prefetches are still 26 function invocations per listing view, which may be worth its own issue, but they make no Analytics Engine queries.

**The layout renders once per navigation, not twice.** The layout markers appear in two flight rows, which is one render streamed in parts.

Caching is not the lever here, for the record: every query already goes through `unstable_cache` (4h for usage, 15min for the admin explorer), keyed on the SQL string. That amortizes *repeat* views of the same prefix, but the first view of each distinct prefix is always a miss — and this feature turned one cache key per product into one per directory. Vercel's Data Cache is also regional and environment-isolated, so the same key can miss several times over, and any change to the SQL (like this PR's) invalidates every entry at once.

## Screenshots

| Product root (unchanged) | Inside a prefix |
| --- | --- |
|  |  |

A path deep enough to outrun the card truncates to the line, with the full path in the title attribute — a wrapped path would push the chart down every time someone navigated somewhere deep:

## Storybook

- [UsagePanel / ScopedToPrefix](https://source-coop-ui-git-feat-analytics-prefix-scope-radiantearth.vercel.app/?path=/story/features-analytics-usagepanel--scoped-to-prefix) — new
- [UsagePanel / ScopedToLongPrefix](https://source-coop-ui-git-feat-analytics-prefix-scope-radiantearth.vercel.app/?path=/story/features-analytics-usagepanel--scoped-to-long-prefix) — new
- [ProductAnalyticsView / UsersUnavailable](https://source-coop-ui-git-feat-analytics-prefix-scope-radiantearth.vercel.app/?path=/story/features-analytics-productanalyticsview--users-unavailable) — new, the nullable `users` state

`UsagePanel`'s `meta` prose covers the new scope.

## Testing

Run against the current head of the branch:

- `npx jest --forceExit` — 748 passed, including `stories.smoke.test.tsx` rendering the three new stories.
- `src/lib/clients/analytics/index.test.ts`: the prefix matches a path and its children but not name-alike siblings; a trailing slash from the URL never reaches the pattern; `%`/`_` in a path are escaped rather than read as LIKE wildcards; `getUsage` sends exactly two queries and mentions `blob8` in neither; the per-IP slices still tile each window with no gap or overlap, now under `getUsageUsers`; and a 429 stops further calls until `Retry-After` has passed.
- `src/components/features/analytics/UsagePanel.test.tsx`: the scope label ends in exactly one slash whether or not the URL carried one, and the product root names no scope at all.
- `npx tsc --noEmit` and lint clean.
- Screenshots above are the two new `UsagePanel` stories rendered from this branch.

## Docs and ADRs

Checked both, no update needed:

- **docs.source.coop** — no page walks through reading a product's analytics; `about-source/core-concepts.md` mentions analytics only as a one-line organization benefit, which still holds.
- **data.source.coop `adrs/`** — 008-analytics records the event schema this reads (`blob3` = `file_path`, truncated to 256 bytes). Querying that column with a prefix rather than an equality moves no decision, and the proxy is untouched.


tylere:
This only appears to work for one level... is that intentional?

Example:

[kerner-lab/fields-of-the-world/austria](https://source-cooperative-git-feat-analytics-prefix-scope-radiantearth.vercel.app/kerner-lab/fields-of-the-world/austria) shows the analytics box

[kerner-lab/fields-of-the-world/austria/label_masks](https://source-cooperative-git-feat-analytics-prefix-scope-radiantearth.vercel.app/kerner-lab/fields-of-the-world/austria/label_masks) does not show the analytics box

alukach:
> This only appears to work for one level... is that intentional?
> 
> Example:
> 
> [kerner-lab/fields-of-the-world/austria](https://source-cooperative-git-feat-analytics-prefix-scope-radiantearth.vercel.app/kerner-lab/fields-of-the-world/austria) shows the analytics box
> 
> [kerner-lab/fields-of-the-world/austria/label_masks](https://source-cooperative-git-feat-analytics-prefix-scope-radiantearth.vercel.app/kerner-lab/fields-of-the-world/austria/label_masks) does not show the analytics box

Good catch, it seems like we're getting rate-limited from Cloudflare

alukach:
This feature seems to have put us up against rate limits. Going to close for now, we can re-approach later when this is more of a pressing need.

Issue #559, opened 2026-09-18
[Bug] User lookup flow is picky


tylere:
**Description of Bug:**

**Steps to Reproduce:**
- Step 1 - Goto https://source.coop/admin/user-lookup and type in part of a name or an entire name (Ex: "nissim")

<img width="870" height="593" alt="Image" src="https://github.com/user-attachments/assets/c75a3f9a-40d7-4197-9151-2ae5e740ff1e" />

- Step 2 - Click on "Look up user". Note the error message "No account found for nissim"

<img width="870" height="593" alt="Image" src="https://github.com/user-attachments/assets/e264c7b8-3aad-4c5c-b7bd-e41f000b1d5e" />

- Step 3 - Return to editing the search text until a list of user accounts appear. 

<img width="870" height="593" alt="Image" src="https://github.com/user-attachments/assets/c4dc8b16-4e85-44a1-81f9-45b222152c6c" />

- Step 4 - Click on the user account you want to lookup. This will replace the search text with the username.

<img width="870" height="593" alt="Image" src="https://github.com/user-attachments/assets/29daaac9-9061-4f18-84b0-42c151551e33" />

Note that the error message "No account for X" is still visible, which is confusing.

- Step 5 - Click on "Look up user". The profile page for the user appears.

<img width="870" height="593" alt="Image" src="https://github.com/user-attachments/assets/450dd5bf-3d47-46ae-9d9b-243dd66ae472" />

**Expected Behavior:**

**Actual Behavior:**

- Despite the instructions saying "Search by username or name, or enter an email address", searching by name doesn't work.
- The error message "No account for X" persists when it is no longer relevant. It should be cleared after selecting a user account from the dropdown list.

**Screenshots:**

**Additional Context:**

alukach:
@tylere I'm not sure we have the right solution in #560, but let's discuss here.

The user lookup form began as being purely "lookup via email", so you could type in an email address and get the respective user. We recently expanded the tool to generate a type-ahead discovery of a user by their account ID or name (#503), this allows you to "lookup via account id".

The first issue you raise is with regards to lookups by user's first/last name. When you type in a user's first name, there's no guarantee that there will only be one user with that username, you need to select the appropriate user in the type-ahead dropdown that appears.

The second issue you raise is with regards the error state on the form not auto-clearing when you change the form input.

[...]

PR #560, opened 2026-09-18
fix(admin): make user-lookup work for name searches and clear stale errors

tylere:
## What I'm changing

Fixes two bugs in the admin user-lookup form (#559):

1. Searching by display name (e.g. "nissim") now works — if the query isn't an exact username match, the action falls back to a name/handle substring search. Exactly one result redirects to that profile; multiple results prompt the user to select from the dropdown.
2. After getting a "no account found" error, selecting an account from the typeahead dropdown now clears the stale error message immediately.

## How I did it

- `lookupUser` in `admin.ts` now calls `accountsTable.searchIndividuals(query)` when `fetchById` returns nothing, handling 0/1/multiple-match cases.
- Added an `onSelect` callback prop to `AccountSearchInput` that fires when the user picks a suggestion.
- `AdminUserLookupForm` uses that callback to increment a `key` on `DynamicForm` (resetting its `useActionState` error) while restoring the selected value via `defaultValue`, so the input keeps the chosen handle.

## How you can test it

1. Go to `/admin/user-lookup`, type a display name (e.g. "nissim") and click "Look up user" — it should redirect to the matching profile instead of showing an error.
2. Type a partial name, get the "multiple accounts match" error, then select from the dropdown — the error should disappear immediately.
3. Type a partial name that matches exactly one user and click "Look up user" directly — should redirect without needing the dropdown.


tylere:
@claude apply changes recommended by the review

tylere:
@alukach added you are a reviewer, because this changes UI behavior.

Issue #547, opened 2026-09-06
Management API and UI for Service Accounts


alukach:
**Phase 2 — The principal.** Required for GitHub Actions (source-cooperative/data.source.coop#223) to be self-serve rather than engineer-configured.

**Today**
None of the Service Account model is visible or manageable.

**Do**
- Create, list, disable and delete under the owner account.
- Add and remove Integrations.
- Add and remove memberships.
- Show when the service account last authenticated.

**Done when**
An organization maintainer can create a service account, grant it a product, and revoke it, without an engineer.

**Follow**
The data-connection admin UI (source-cooperative/source.coop#364) is the template: list, create and edit pages under `edit/account/[account_id]/<resource>/`, tab-wired in that route's `layout.tsx`, gated by `canManageAccountDataConnections`, with a platform-admin twin under `admin/`.

**Watch**
- The creator must hold a qualifying membership role on the owner account.
- That gate reaches `canManageAccount` → `hasRole` and its self-authorization shortcut, so these pages sit inside source-cooperative/source.coop#544's blast radius. Land it first.
- Deleting an account that owns service accounts should be blocked; a missing owner must deny rather than skip the check. Account deletion itself is source-cooperative/source.coop#355 and unimplemented, so the guard attaches there.

Depends on source-cooperative/source.coop#545 and source-cooperative/source.coop#546.

---

Part of source-cooperative/source.coop#491.

Issue #545, opened 2026-09-06
Add the Service Account account type


alukach:
**Phase 2 — The principal.**

**Today**
Two account types, individual and organization. 48 places across 19 files branch on which — 77 across 31 including tests, stories and mocks. `AccountType.SERVICE` already exists as a commented-out stub in `src/types/account.ts`, as does a commented-out authz branch, and the members route's OpenAPI text already describes service-account invite behaviour.

**Do**
- Add `AccountType.SERVICE`.
- Add a persisted `owner_account_id` and an index for it. Today the field exists only on the organization *creation request*, consumed once to seed an `Owners` membership; nothing on the account records ownership.
- Restrict a service account to `read_data` / `write_data` memberships — never `owners` or `maintainers`.
- No reserved id namespace. `type` is the discriminator and `owner_account_id` records ownership; [`ID_REGEX`](https://github.com/source-cooperative/source.coop/blob/b77a514b1a68cc4c642e33b50d84379b16c88014/src/types/shared.ts#L25-L30) stays as it is.

**Done when**
- A service account can be created.
- It appears on the two membership-management pages, visually distinguished, since that is where an owner revokes its grant.
- It cannot be invited as owner or maintainer, cannot create products or organizations, and cannot hold the admin flag.

**Watch**
- The default at every branch site is **exclude**; inclusion needs a reason.
- Most of those sites are the binary `isIndividualAccount` / `isOrganizationalAccount` helpers, which would render a machine as an *organization profile* rather than failing.
- `AccountSchema` is a `z.discriminatedUnion("type", …)` where `identity_id` is a required string on the individual arm and `z.undefined()` on the organizational arm. A service account with N bindings and no single `identity_id` fits neither — a schema change, not an enum addition.
- Two gates block memberships outright until changed: [`memberships.ts`](https://github.com/source-cooperative/source.coop/blob/b77a514b1a68cc4c642e33b50d84379b16c88014/src/lib/actions/memberships.ts#L79) rejects any invite where the invited account is not an individual, and the account memberships page redirects unless the target is an organization.
- An account id is also the first segment of a public URL, so each service account consumes a name and gets a profile route.

**Open question carried from the epic:** what "capped by the owner" means when the owner is an organization — which is also the recommended default. Organizations never authenticate, and the membership lookup returns rows where the organization is the *member*, which is not the same thing.

Depends on source-cooperative/source.coop#543 and source-cooperative/source.coop#544.

---

[...]

Issue #543, opened 2026-09-06
Resolve accounts by (issuer, subject)


alukach:
**Phase 1 — Foundations.** Gates the Integrations model, the management UI and subject namespacing. This is the epic's schedule risk.

**Today**
A token is turned into an account by looking up its `sub` claim alone ([`oidc.ts`](https://github.com/source-cooperative/source.coop/blob/b77a514b1a68cc4c642e33b50d84379b16c88014/src/lib/api/oidc.ts#L153)), and that lookup then filters to individual accounts. So no other account type can authenticate at all, and two issuers using the same subject string would be indistinguishable. The issuer is already pinned — to the data proxy, not Ory — so this is a subject-namespace problem, not an unvalidated-issuer one.

**Do**
- Add a table of identity bindings: `(issuer, subject) -> account`.
- Make the account lookup take both claims, for identities the platform does not own (the proxy's own issuer for API keys, GitHub Actions). The existing single-argument lookup stays as it is: an individual's Ory identity lives on the account row as `identity_id`, and is not a binding.
- Enforce uniqueness on the pair — two rows claiming the same `(issuer, subject)` is a hard error, not last-writer-wins.

**Decided 2026-09-22 (PR #565):** no backfill of Ory identities, no dual-read, no index retirement. `identity_id` can't leave the row (session, email lookup, proxy credentials read it), so a backfill would be a second copy of the same fact; the Ory "issuer" would be a config value no token carries; and uniqueness only matters for attacker-chosen subjects, which an Ory subject never is. The bindings table holds only other issuers.

**Done when**
- The lookup takes `(issuer, subject)` for non-Ory issuers, uniqueness enforced by the key.
- Deleting an account removes its bindings.
- An Ory identity still resolves through `identity_id`, untouched.

**Watch**
- The individual-only filter is applied in **two** places: the data layer and [`getPageSession`](https://github.com/source-cooperative/source.coop/blob/b77a514b1a68cc4c642e33b50d84379b16c88014/src/lib/api/utils.ts#L106). Fixing only the data layer leaves cookie sessions broken for any new account type.
- A new table is declared in **four** places, not two: `deploy/lib/database-construct.ts`, `scripts/init-local.ts`, `fixtures/`, and `src/lib/api/utils.mock.ts` (which `init-local.ts` imports for seed data).
- A binding's issuer must never be an empty string from unset config; the table client refuses one.

---

Part of source-cooperative/source.coop#491.

PR #563, opened 2026-09-21
feat(accounts): add the Service Account type

alukach:
First half of #545, stacked on #562 (merge that first; this PR's base is its branch). Part of #491.

## What

A third account type, `SERVICE`: a machine principal owned by another account.

**Model.** `ServiceAccountSchema` is a new arm of the discriminated union — `identity_id: z.undefined()` like an organization (it authenticates through identity bindings, #543/#546), plus a required `owner_account_id`. That answers the schema question the epic flagged: a service account carries no single identity id, so it fits neither existing arm, and now has its own. No reserved id namespace — `type` is the discriminator. `ServiceAccountCreationRequestSchema` mirrors the organization one.

**Storage.** `owner_account_id` gets an index, declared in all four places a new attribute has to be: `deploy/lib/database-construct.ts`, `scripts/init-local.ts`, `fixtures/`, and `utils.mock.ts` (which builds sessions for service accounts now, since they are principals). `AccountsTable.listByOwner` queries it; `isServiceAccount` joins the other guards.

**What it may do**, per the epic:
- `isAdmin` returns false for a service account whatever its flags say, and both flag write paths (`updateAccountFlags`, `PUT …/flags`) refuse to store `ADMIN` on one.
- It creates neither products (`createRepository`) nor accounts (`createAccount`) — tested with the flags that would otherwise allow it.
- It holds only `read_data` / `write_data`, only on products its owner owns, never organization-wide. `serviceAccountGrantProblem` states that rule once; the invite action and all three membership routes (account members, product members, update-role) apply it. The product-members route had no type gate at all before — it accepted organizations.
- Its owner grants access directly as a `Member`: there is nobody at the keyboard to accept an invitation.
- It holds no rights over itself (#562). Whoever manages its owner — `Owners`/`Maintainers` on the owner, or the individual owner themselves — creates, edits and disables it.
- It has no public profile: `/[account_id]`, its metadata and its `og` preview image return not-found, `GetAccountProfile` is refused for everyone (admins included; the profile API already admits only individuals), and `AccountSchemaMetadata` emits nothing for it.

`createAccount` accepts `type=service` with an `owner_account_id` and redirects to the owner's page.

**Not in this PR:** rendering service accounts on the two membership-management pages, visually distinguished (that's the second half of #545, next in the stack), and a creation form (#547).

## Testing

- `npx jest` — 69 suites, 754 tests pass. New: `service accounts` block in `authz.test.ts` (never admin even when flagged; creates nothing even when flagged; created/edited/disabled by the owner's owners and maintainers, not readers; a person may create one under themselves; no rights over itself; no profile for anyone, admins included; writes only the product it's granted, via a `write_data` membership fixture) and an `inviteMember` block in `memberships.test.ts` (grants read_data on the owner's product directly as a member; refuses owners/maintainers; refuses another owner's product and organization-wide grants; refuses an organization as invitee; still invites a person).
- `npx tsc --noEmit` — clean. The third union arm surfaced exactly one site, `AccountSchemaMetadata`, now handled.
- `next lint` on the changed files — clean (two pre-existing `any` warnings in `accounts.ts`, untouched).

## Docs and ADRs

No component changes, so no stories; the membership-page rendering lands with its stories in the next PR. docs.source.coop: no user flow changes yet — creation is API/action-only until #547; the unattended-workflow page is source-cooperative/docs.source.coop#34. data.source.coop ADRs: the Service Account decision itself is source-cooperative/data.source.coop#230, which the epic sequences before this step; this PR implements the model as #491 specifies it and does not move a recorded decision.


alukach:
Dropped the unreachable `if (!accountToUpdate)` block. Kept `ServiceAccountCreationRequestSchema` as its own name — the two schemas are the same shape today, but a reader looking for what a service account is created from should find it under that name. `listByOwner` gets its caller in #547's management UI.

alukach:
Pushed 056d8f94 for the og/profile note: `GetAccountProfile` is now refused for a service account for everyone, admins included, and the `og` route answers "Account Not Found" for one, matching the page. The profile API route already admitted only individuals, so it was closed; the authz helper behind it is now closed too. The description's deferral paragraph is gone and its profile bullet says what the branch does now.

alukach:
Pushed a sweep for the type-guard note: every place that compared an account's type to `SERVICE` now asks `isServiceAccount`, in authz, the actions, the routes, the page and the og image. The three guards moved to `src/types/account.ts` so authz can import them without the database client; the database module re-exports them, so nothing else changed. The two sites left are on the creation request's `type` string, not an account. Description still holds.

PR #564, opened 2026-09-21
feat(memberships): show and grant service accounts on the membership pages

alukach:
Second half of #545 — closes it. Stacked on #563 (which is stacked on #562); merge those first. Part of #491.

## What

Service accounts now appear on the two membership-management pages, visually distinguished, and can be granted access from them.

- **`MembershipsTable`** — a member that is a service account is badged "Service account", and its name is not a link: a machine has no profile page to land on (#563 made `/[account_id]` 404 for it). Everything else about the row is unchanged.
- **Invite picker** — `InviteMemberForm` passes the account being managed to `AccountSearchInput` as `memberOf`, and the search offers that account's service accounts alongside people. `searchAccounts(query, memberOf?)` → `AccountsTable.searchMemberCandidates` (was `searchIndividuals`): same scan, one more eligibility clause, `owner_account_id` added to the projection. Without `memberOf` it behaves exactly as before. A suggestion carries the account's `type`, and a service account is badged "Service account" in the list, the same badge the table gives it.
- The field is labelled "Account" rather than "User" now that it can hold either.

Not done: hiding Owner/Maintainer in the role select when a service account is picked — the action already refuses with a clear message (#563), and the form doesn't know the picked account's type.

## Stories

On this branch's deploy:

- `MembershipsTable` › **WithServiceAccount** — https://source-coop-ui-git-feat-service-account-mem-547052-radiantearth.vercel.app/?path=/story/features-memberships-membershipstable--with-service-account
- `InviteMemberForm` › **Default** (type `mis` — the mock returns a service account for `miskatonic`) — https://source-coop-ui-git-feat-service-account-mem-547052-radiantearth.vercel.app/?path=/story/features-memberships-invitememberform--default
- `AccountSearchInput` › **OfferingServiceAccounts** — https://source-coop-ui-git-feat-service-account-mem-547052-radiantearth.vercel.app/?path=/story/components-forms-accountsearchinput--offering-service-accounts

## Screenshots

`MembershipsTable` with a service account among the members — badged, unlinked, sorted below the owners and maintainers:

A memberships table: Sam Rivers (Reader, Invited), Alice Coltrane (Owner, Member), and Nightly Sync with a "Service account" badge (Writer, Member)

The picker offering the organization's service account alongside people, badged:

The account picker with "mis" typed, suggesting Alice Coltrane, Miskatonic University, and Miskatonic Bot with a "Service account" badge

## Testing

[...]

PR #565, opened 2026-09-21
feat(accounts): resolve accounts by (issuer, subject)

### LINKED ISSUE #543: Resolve accounts by (issuer, subject)

alukach:
Closes #543. Independent of stack #571 (#564 → #566 → #567 → #570); based on `main`, and #566 carries these commits until this merges. Part of #491.

## What

An account becomes findable by *how it signed in*, for identities the platform does not own: a service account's subject under the data proxy's issuer (#570), a GitHub Actions workflow (#566), and whatever issuer comes next.

**The table.** `identity-bindings`, keyed `(issuer, subject)` → `account_id`, with an `account_id` index for listing an account's bindings. The composite key *is* the uniqueness rule the issue asks for: a subject binds to one account per issuer and nothing more, enforced by DynamoDB's conditional write rather than a check, and surfaced as `IdentityAlreadyBoundError`. Declared in the places a table has to be — `deploy/lib/database-construct.ts`, the `api-stack` grant list, `scripts/init-local.ts`. Both halves of the key must be non-empty: an issuer read from unset config would be `""`, and every binding under it would share one partition.

**The lookup.** `AccountsTable.fetchByIdentity(issuer, subject)` resolves any account type through a binding, and `AccountsTable.delete` removes an account's bindings with it, so an orphaned pair can't keep a subject from ever binding again.

**Ory identities stay where they are — a decision, flagged.** #543 asked for every individual to be backfilled into the table under an Ory issuer, with a dual-read window and the `identity_id` index retired after. Three reviews of that plan reached the same answer, and this PR takes it: an individual's Ory identity is not a binding. `identity_id` can't leave the row (the session cookie, the Ory email lookup, proxy credentials and key minting all read it), so a backfill would have made a second copy of the same fact to keep in step, in exchange for dropping one index. The "issuer" it would be keyed under is the Ory SDK URL from config, which no token carries as `iss`; a change to that variable would have orphaned every person's login row at once. And the conditional write only buys something when the subject is attacker-chosen, which an Ory subject never is (it comes from the verified session, and authz already refuses a second individual per principal). So `fetchByOryId` is untouched, `create` writes no binding, there is no backfill, no dual-read and no follow-up. The OIDC resolver's chain, `fetchByOryId(sub) ?? fetchByIdentity(proxyIssuer, sub)` in #570, already dispatches between the two. If a person ever needs several identities, or the platform leaves Ory, bindings under a real `iss` are the migration then.

## Testing

- `npx jest` — all suites pass. `identity-bindings.test.ts`: conditional write, the conflict named as `IdentityAlreadyBoundError`, resolution by key, and an empty issuer or subject refused before the table is touched. `accounts.identity.test.ts`: an Ory identity resolves through the index and never reaches the bindings table; `fetchByIdentity` resolves any type while `fetchByOryId` stays individual-only; deletion removes bindings; issuers are told apart. `accounts.create.test.ts` is unchanged from `main`.
- `npm run type-check` — clean (`scripts/` is in the project).
- `next lint` — clean apart from pre-existing warnings in `accounts.ts` and `init-local.ts`.

## Docs and ADRs

docs.source.coop: no user-facing flow changes — sign-in works as before. data.source.coop: ADR-014 (source-cooperative/data.source.coop#232) said every individual is bound under the Ory issuer and listed the dual-read window as a cost; that PR is amended to say Ory identities stay on the account row and the table holds only other issuers. source-cooperative/data.source.coop#222 (issuer-qualified subjects) still has `fetchByIdentity` to resolve against, with one branch for the Ory issuer. #543's "Do" and "Done when" are updated to match.


alukach:
Addressed the delete finding: `AccountsTable.delete` now removes the account's bindings first, with a test. Left the rollback as-is on purpose — removing the row on *any* binding failure is the safer side: no account may exist without a binding, whatever went wrong writing it.

alukach:
Both addressed. `IdentityBindingsTable.create` now (a) returns without writing when the table doesn't exist yet — the `identity_id` fallback carries signups until the backfill, so either-order deploys hold for writes too — and (b) throws a named `IdentityAlreadyBoundError` on a pair conflict, which `createAccount` reports as "This sign-in already has an account" rather than "account ID taken". Tests cover the missing-table signup, the conversion, and the rollback's new error name.

alukach:
Rebased onto `main` and pushed a commit that takes the no-backfill outcome from the three-way review: `fetchByOryId` is back to the `identity_id` index, `create` writes no Ory binding, and the backfill script, dual-read fallback and missing-table tolerance are gone. The bindings client now refuses an empty issuer or subject. Description rewritten; #543 and ADR-014 (source-cooperative/data.source.coop#232) updated to match.

alukach:
Pushed 788c22d7 for the code-review findings: `AccountsTable.delete` takes the account id alone (the key it sent carried a `type` attribute the table isn't keyed on, so DynamoDB would have rejected it after the bindings were already gone), the bindings and the row now go in one `TransactWriteCommand`, the bindings are listed past the request cache, `create` validates the row with `IdentityBindingSchema` instead of a duplicate check, the test fake no longer removes the last binding on a miss, and the schema's OpenAPI example shows a GitHub Actions binding. Description still holds.

Issue #546, opened 2026-09-06
Integrations model for Service Accounts


alukach:
**Phase 2 — The principal.**

**Today**
Nothing records how an account may authenticate. There is one implicit route, and it is assumed to be Ory.

**Do**
- Let a service account hold many identity bindings, each naming an issuer and an **exact** subject (no prefixes, no patterns).
- Require proof of control of the subject when attaching one.
- Support two Integration types to start: **API Key** and **GitHub**.

**Done when**
- A service account with two Integrations authenticates by either and gets identical access.
- Attaching a subject the user cannot prove control of is rejected.

**Why proof matters**
Without it, anyone could claim another organization's CI subject and receive their access. For GitHub the proof is a token minted by the workflow itself carrying a one-time audience. An issuer that cannot mint on demand needs reviewed registration instead.

**Note:** permissions live on the service account, not on the Integration. Both routes reach the same account with the same access. If CI should have different access than a key, that is two service accounts.

Depends on source-cooperative/source.coop#543 and source-cooperative/source.coop#545.

---

Part of source-cooperative/source.coop#491.

Issue #223, opened 2026-02-10
[Bug] UI doesn't support paths that coincide with a filename


alukach:
**Description of Bug:**

Consider a product containing a file has the same name as a prefix, such as:

```
aws s3 ls s3://my_account/my_product/ --recursive
2026-02-07 09:21:27       9605 foo.xyz
2026-02-09 11:30:39       9638 foo.xyz/bar.md
```

When a user navigates to `source.coop/my_account/my_product/foo.xyz` they will render a preview of the `s3://my_account/my_product/foo.xyz` file, disregarding the existence of any file that may have `foo.xyz/*` as part of its prefix e.g. `foo.xyz/bar.md`.

**Expected Behavior:**

The UI viewer should also check for any files with descendant prefixes when viewing a path. Possibly rendering both the file navigator and the file preview.

**Actual Behavior:**

**Screenshots:**

**Additional Context:**

See #222

Issue #222, opened 2026-02-10
[Bug] Web Interface Upload - Update README.md Not Displayed Unless Click on README.md


diegoripley:
**Description of Bug:**
The `README.md` on https://source.coop/dataforcanada/ca-orthoimagery-labs is an older version of the one at https://source.coop/dataforcanada/ca-orthoimagery-labs/README.md

**Steps to Reproduce:**
- Step 1
Open https://source.coop/dataforcanada/ca-orthoimagery-labs
- Step 2
Look at the table listed in the "Download and Preview" section. It will not have the PMTiles Torrent "Download" URLs  

**Expected Behavior:**
Open https://source.coop/dataforcanada/ca-orthoimagery-labs/README.md , it shows the expected `README.md` with the updated "Download" URLs.

**Actual Behavior:**
The `README.md` listed in https://source.coop/dataforcanada/ca-orthoimagery-labs is not the same one listed in https://source.coop/dataforcanada/ca-orthoimagery-labs/README.md . It is most definitely an older version of the file.

**Screenshots:**

Actual Behavior

<img width="1567" height="1707" alt="Image" src="https://github.com/user-attachments/assets/ae799d21-2748-44ae-8cb0-71a9b652b034" />

Expected Behavior

<img width="1608" height="1783" alt="Image" src="https://github.com/user-attachments/assets/a9bc0e80-ad61-4aca-a43f-ef2d3092557a" />

**Additional Context:**
- I uploaded the initial `README.md` via rclone
  - The newer `README.md` (expected) was uploaded via the web interface upload 

Even with the following, I encounter the same behaviour:
- Cleared browser cache
- Cleared DNS cache
- Tried with various other browsers

alukach:
This is a weird bug. Looking at the product, we see that the `README.md` object is marked as a directory (ie has a `>` symbol):


<img width="180" height="156" alt="Image" src="https://github.com/user-attachments/assets/bad79a32-fea7-491e-b0c5-15a43e626e09" />


Looking at the data, we see that you have two READMEs in your product:

```
▶ aws s3 ls s3://us-west-2.opendata.source.coop/dataforcanada/ca-orthoimagery-labs/README --recursive
2026-02-07 09:21:27       9605 dataforcanada/ca-orthoimagery-labs/README.md
2026-02-09 11:30:39       9638 dataforcanada/ca-orthoimagery-labs/README.md/README.md
```

[...]

Issue #221, opened 2026-02-09
[Bug] image preview - broken if the file is too big?


severo:
**Description of Bug:**

The image preview shows a black background. I'm not sure if it's normal, but I'm reporting it just in case.

**Steps to Reproduce:**

https://source.coop/clarkcga/multi-temporal-crop-classification/hls/chip_002_060_merged.tif

The image is black, and the console shows errors:

<img width="855" height="208" alt="Image" src="https://github.com/user-attachments/assets/3c49ff94-1a8a-40ab-ac0c-c6f731b12fe1" />

**Expected Behavior:**

Show the image, or a clear error message.

**Actual Behavior:**

A black background.

**Screenshots:**

<img width="969" height="746" alt="Image" src="https://github.com/user-attachments/assets/4c339f08-6160-42fe-9ed4-1e4b01c6187a" />

**Additional Context:**

jedsundwall:
@alukach mentioned an approach at some point yesterday in which we could add some logic to load different TIFF viewer frames based on what we can tell about the TIFF. We now do this to detect certain kinds of Parquet files: https://github.com/source-cooperative/source.coop/blob/48de05030fe7fc06647457535ee249389cc237fc/src/components/features/products/object-browser/ObjectPreview.tsx#L11 

In the meantime, I think we should remove TIFF support. My intuition is to display nothing rather than be forced to throw errors that might make people mistakenly think that there's something wrong with their COGs.

PR #566, opened 2026-09-21
feat(accounts): trust subjects per account, the way a role's trust policy does

alukach:
Stacked on `main` (#564 and #565 have merged); the bottom of stack #571. Part of #491; the model half of #546 — the GitHub-specific half is #567.

## What

A service account says which subjects may act as it, the way an AWS role's trust policy does. That replaces the identity-bindings model #565 merged, before anything was written to it. **A decision from review, flagged:** the old model resolved a subject alone to an account, which meant a subject could be squatted and every binding needed proof of control — a signed challenge, a one-off workflow run, a completion endpoint. Under this model the workload names the account it wants when it exchanges its token, so trusting a subject one doesn't control gains nothing, and there is nothing to prove first. It is also the flow people already know from integrating GitHub Actions with AWS.

- **`account-trusts` table**, keyed by the account: partition `account_id`, sort `identity` (`issuer` + space + `subject`). "Does *this* account trust *this* subject" is one exact read; an account's trusts are one query; a subject may be trusted by any number of accounts. Replaces `identity-bindings` in the CDK construct, the `api-stack` grant and `init-local`. The merged table is empty and was never written to, so this is a rename, not a migration; the old table is retained by policy and can be deleted by hand.
- **`AccountTrustsTable`**: `isTrusted`, and `listByAccount` (paged, with `bypassCache` for destructive callers). `AccountsTable.delete` removes an account and its trusts together, in transactions of at most 100 with the row last. Writing and removing a trust arrive with #567, where the settings page calls them; nothing here writes one.
- **The exchange check**: `POST /api/v1/accounts/{id}/trusts/exchanges` with `{ issuer, subject }`, authenticated as the account. This is what the data proxy asks at `/.sts` after verifying a platform IdP's token and reading the account from `RoleArn` — the trust-policy check of an assume-role call. The only route here, and the proxy is its only caller.
- **Resolution**: a proxy-signed subject that is a service account's own id resolves to it; a person's Ory identity is read alongside, and a subject that names both is refused as ambiguous. A person's or organization's handle never resolves. The exchange route needs this to authenticate the proxy as the account.
- The GitHub issuer constant and the subject pin — one repository, one ref or environment, nothing organization-wide — arrive with #567, where the trusts are written; nothing here reads them.

**Removed from `main`:** the `identity-bindings` table, its client and type, and `AccountsTable.fetchByIdentity`, none of which had a caller. Nothing in this app verifies a platform IdP's token: the proxy does, as it already does for every issuer it trusts (source-cooperative/data.source.coop#223).

**The ARN** a workflow uses: `arn:aws:iam::<service-account-id>:role/FullAccess` (or `ReadOnly`, or the `_default` alias) — the account where AWS puts the account number, the ceiling where AWS puts the role name. The partition is `aws` because `aws-actions/configure-aws-credentials` treats any other value as a bare role name; SDKs check only the value's length. The proxy already ignores the partition segment and now reads the account one (source-cooperative/data.source.coop#221, #222).

## Testing

- `npx jest` — all suites pass. `account-trusts.test.ts`: one exact read; cache bypass; a paged listing followed to the end. `accounts.trusts.test.ts`: Ory resolves through the index and never touches trusts; delete is one transaction over trusts and the row, and 150 trusts split into two with the row last. `trusts/exchanges/route.test.ts`: trusted, not trusted, an admin asking about any account, 401 for another account, 400 without both fields. `oidc.test.ts`: a service account resolves by its own id after Ory; a handle never does; a subject naming both a person and a service account is refused.
- `npm run type-check`, `next lint` — clean.

## Docs and ADRs

data.source.coop: ADR-014 (source-cooperative/data.source.coop#232) is amended to this model — trusts per account, no proof of control, `RoleArn` names the account, the exchange route. #221 (`RoleArn` account segment), #222 (what the proxy forwards: issuer, subject, and the account named) and #223 (GitHub as a trusted issuer, checked through the route) carry the contract as comments. docs.source.coop: the unattended-workflow guide (source-cooperative/docs.source.coop#34) will carry the workflow step #567 issues; nothing existing describes this.


alukach:
Pushed a follow-up so the comparisons this branch added ask `isServiceAccount` (the guard now lives in `src/types/account.ts`, from #563). Description still holds.

alukach:
Rebased: this branch now carries #565's no-backfill commit (`fetchByOryId` back to the `identity_id` index, no Ory binding on create, no backfill script). Nothing in the GitHub-integration diff changes; the description still holds.

alukach:
Rebased onto `main` now that #564 has merged (its commits are dropped from this branch), and carrying #565's latest commit. The GitHub-integration diff itself is unchanged.

alukach:
Rebased onto `main` past #565's merge (this branch is now just its own commits), and pushed 74f020f1 for the review finding. The disabled check went into the two integration routes rather than into `canManageServiceAccount`: that function is also what authorizes re-enabling and deleting a service account in #567, so refusing disabled accounts there would have locked a paused account for everyone but an admin. Both the challenge and the completion route now answer 403 for a disabled account, the second so a challenge issued before the account was disabled can't complete after; tests cover both. The doc comment now carries the admin caveat. #570 gets the matching rule for keys. Description still holds.

alukach:
Rewritten on the trust model: an `account-trusts` table keyed by the account replaces identity bindings, proof of control is gone, and a workflow names the account in `AWS_ROLE_ARN` (`arn:aws:iam::<service-account-id>:role/FullAccess`). Stack #571 is rebuilt as `main ← #566 ← #567 ← #570`, one commit each; descriptions rewritten. ADR-014 (source-cooperative/data.source.coop#232) amended to match.

alukach:
On the review: the description and title were rewritten for the trust model right after that run (the review ran against the push, before the rewrite landed). Two follow-ups from its ponytail notes: `GITHUB_ACTIONS_ISSUER` and the subject regex had no caller here, so they move to #567 where trusts are written; and this PR is now one commit. `AccountTrustsTable.create`/`delete` stay, as the staged surface #567 calls.

alukach:
Pushed fixes from a `/code-review` pass: the delete cascade now pages through an account's trusts and splits the deletes into transactions of at most 100, the account row last, so a failure part-way never leaves trusts without an account; `canManageServiceAccount` checks the account type before the admin shortcut, so it is false for an admin over an organization too (test added); the resolver reads the Ory and service-account namespaces together and refuses a subject that names both, which is the invariant the comment used to promise a stacked PR would enforce; `isTrusted` logs like the other table methods; the sort-key comment no longer claims subjects have no whitespace, and #567's subject regex now admits environment names with spaces. Description still holds.

alukach:
Follow-up on unused code: `canManageServiceAccount`, `AccountTrustsTable.create`/`delete` and `AlreadyTrustedError` had no caller here either, so they and their tests move to #567 with the settings actions that use them. What stays is what this PR's own code calls: `isTrusted` (the exchange route), `listByAccount` and `identityKey` (the delete cascade), `serviceAccountById` (the resolver). Description updated.

alukach:
On the review: it ran between the push that moved `create`/`delete`/`AlreadyTrustedError` to #567 and the description update that followed, so the mismatch it saw was the description lagging the code by a few minutes; the current description lists only what this diff calls. Its one non-blocking note is taken: the exchange route's admin path now has a test. CI's type-check and tests are the confirmation it asked for.

PR #568, opened 2026-09-21
fix(admin): search users as you type and list every match

### LINKED ISSUE #559: [Bug] User lookup flow is picky

alukach:
Closes #559.
Closes #560.

https://github.com/user-attachments/assets/0ce839c7-3063-4e28-8be2-8d8b6613fe03

## What changed

The admin user lookup submitted a single handle to a server action that redirected to that profile. Searching by name ("nissim") failed with "No account found" unless the admin first picked a suggestion from the type-ahead, and the stale error stayed on screen after they did.

Typing is now debounced into the URL's `q` param, the page runs the search on the server, and it renders one card per match, each linking to the profile. There is no submit button. The URL is the single source of truth, so a result page can be reloaded, shared, or returned to with the back button. From the first keystroke that differs from the URL until the new page renders, the field shows a spinner and the results beneath it are dimmed, so the old answer is not read as the new one. Under the search box a line says which system answered: an email is resolved through Ory to one identity, and anything else is matched against handles and display names in the database. An email Ory knows but that has no source.coop profile is told apart from one Ory has never seen. Disabled accounts are included and flagged with a badge, since an admin is as likely to be looking for one of those as not.

- `src/app/(app)/admin/user-lookup/page.tsx` reads the query, calls `searchUsers`, and passes plain values to the presentational component.
- `src/components/features/admin/AdminUserSearchField.tsx` is the client half: a text field debounced into `router.replace`, owning the in-flight state and wrapping the results so it can dim them.
- `src/components/features/admin/AdminUserLookup.tsx` renders the field, the source line, and the result cards. It replaces `AdminUserLookupForm`.
- `src/lib/api/user-lookup.ts` replaces the `lookupUser` server action and returns a `UserSearch` that says where it went.
- `searchMemberCandidates` takes an options object for `limit` and `includeDisabled`; the membership pickers are unchanged.

## Storybook

- [Empty](https://source-coop-ui-git-fix-admin-user-lookup-results-radiantearth.vercel.app/?path=/story/features-admin-adminuserlookup--empty)
- [DatabaseMatches](https://source-coop-ui-git-fix-admin-user-lookup-results-radiantearth.vercel.app/?path=/story/features-admin-adminuserlookup--database-matches)
- [Searching](https://source-coop-ui-git-fix-admin-user-lookup-results-radiantearth.vercel.app/?path=/story/features-admin-adminuserlookup--searching), driven by a play function that types into the field
- [DatabaseNoMatches](https://source-coop-ui-git-fix-admin-user-lookup-results-radiantearth.vercel.app/?path=/story/features-admin-adminuserlookup--database-no-matches)
- [OryMatch](https://source-coop-ui-git-fix-admin-user-lookup-results-radiantearth.vercel.app/?path=/story/features-admin-adminuserlookup--ory-match)
- [OryIdentityWithoutProfile](https://source-coop-ui-git-fix-admin-user-lookup-results-radiantearth.vercel.app/?path=/story/features-admin-adminuserlookup--ory-identity-without-profile)
- [OryNoIdentity](https://source-coop-ui-git-fix-admin-user-lookup-results-radiantearth.vercel.app/?path=/story/features-admin-adminuserlookup--ory-no-identity)

Screenshot to follow before this leaves draft.

## Docs and ADRs

Checked docs.source.coop: no page describes the admin user lookup, so nothing there dates. The data-proxy ADRs don't cover admin tooling.

## Testing

`npm run type-check`, `npm run lint`, and the full `jest` suite pass on the head commit, including the story smoke test for the seven stories above. Unit tests cover `searchUsers` (email via Ory, unknown email vs identity without profile, name search, blank query) and the `includeDisabled` option on `searchMemberCandidates`. The branch is rebased onto main after #562–#564; both Vercel previews built on the previous head and are rebuilding on this one.


alukach:
Addressed in the latest commit. The two triggers are now separate effects sharing a ref of the query last seen in the URL, per the suggested shape: one resyncs the field when the URL changes from outside, the other pushes a settled edit into the URL and is keyed only off the debounced value. `AdminUserSearchField.test.tsx` covers the Back case (edit, navigate, re-render with the previous query, assert the field follows and `router.replace` is not called); it fails against the previous component and passes now.

PR #569, opened 2026-09-21
fix(api): accept every membership role when updating one

alukach:
Independent of the service-account stack, but it bites there: `PUT /api/v1/memberships/{id}/update-role` validated the requested role against `["owners", "maintainers", "members"]`. `"members"` is not a `MembershipRole`, and `read_data` and `write_data` — the only two roles a service account may hold (#563) — were refused with `Invalid role`. The route now validates against the enum. Part of #491.

## Testing

New `route.test.ts`: every `MembershipRole` value is accepted and written; `"members"` and `"admin"` are refused without a write. `npm run type-check` clean.

## Docs and ADRs

No user-facing flow or recorded decision changes. The route's OpenAPI enum now lists the four real roles (it said `members`, which never existed); nothing in docs.source.coop describes this route, and the data-proxy ADRs don't cover membership roles.


alukach:
Pushed the OpenAPI enum fix from the review; the Docs and ADRs note now says what the branch does. Title and description otherwise hold.

PR #567, opened 2026-09-21
feat(accounts): manage service accounts from account settings

alukach:
Closes #547. Part of #491. Stacked on #566, now merged, so the diff is this branch alone: the feature commit, the immutable-subject commit, two review-fix commits, a redesign of the form and the list, a fix for writing trusts, and owner-namespaced ids.

## What

The real version of the flow #495 mocked, built on the model (#563), the membership pages (#564) and account trusts (#566). Rebuilt on the trust model in one commit; what review changed follows as commits of its own.

**Settings → Service Accounts** appears for individuals and organizations alike, to whoever `canManageAccount` says manages the account. The list page loads each service account with its trusts, its grants and (from #570) its keys.

**`ServiceAccountForm`** — who it is (the name, and the id derived from it, shown as code with an Edit link the way the data-connection form shows its id), how software signs in (any number of GitHub workflows, each a card with the repository on its own row — `owner/repo`, or `owner@id/repo@id` for the immutable subjects GitHub mints for repositories created after July 2026 or opted in — then a Ref/Environment segmented control and its value, and the exact subject shown as you type), what it may reach (each of the owner's products as a row with a None / Read / Read and write control). **The id is namespaced under the owner:** `{owner}--{id}`, such as `acme--nightly-sync`, joined by the `--` account-owned data connections already use. Two owners can each have a `nightly-sync`, a service account never takes a handle a person or organization might want (their ids cannot contain `--`), and no service-account id can equal an Ory identity id, which is a UUID. The form asks for the short id and shows the whole one; the action composes it. `ServiceAccountSchema` requires the composed form, and a membership's member id accepts it, which granting a namespaced service account a product needs. The general `createAccount` action no longer takes `type=service`: nothing posted it, and it created service accounts without the owner checks or the namespace. Submitting writes the account, its grants and its trusts, and shows **`ServiceAccountCreated`**: the account is ready, and each trusted workflow gets the step it adds to act as it. Nothing has to run first, and the step carries nothing that expires.

**`ServiceAccountList`** — one row per account in the data connections' own `ConnectionList`/`ConnectionRow`: the name linking to the account's page, its id, a marker when it is disabled, and how many workflows it trusts and products it reaches.

[...]

Issue #561, opened 2026-09-21
Register the API-key marker with secret scanners and expose self-revoke


alukach:
**Alongside.** Starts on day one; the only item in the epic on someone else's clock.

**Do**
- Choose a distinctive marker for ADR-013 API keys. A JWT begins `eyJ` like every other JWT, so scanner registration needs something to match — a string prefix the proxy strips before verification, or a claim-based pattern. The choice belongs in source-cooperative/data.source.coop#230.
- Register it with secret-scanning partners (GitHub, and others as relevant) and implement their verification endpoint.
- Expose a public self-revoke endpoint so a stranger who finds a leaked key can kill it without holding rights on the account.

**Why now**
Partner registration is a multi-week external process. The marker must be fixed before source-cooperative/source.coop#548's schema. The legacy prefix was a bare `SC` (`src/lib/actions/crypto.ts`), far too low-entropy to register.

Blocks source-cooperative/source.coop#548.

---

Part of source-cooperative/source.coop#491.

Issue #548, opened 2026-09-06
API key lifecycle for Service Accounts


alukach:
**Phase 5 — API keys.**

**Today**
No key mechanism exists. The legacy `api-keys` table is unrelated and is no longer honoured by the proxy.

**Do**
- Issue a key: shown once, bound to exactly one service account.
- Store `expires_at` (nullable) on the key record, **not** in the token. Default to a bounded expiry; allow "never expires" as an explicit choice with a visible warning.
- List, revoke, rotate, and record last-used.
- Allow expiry to be changed after issuance — extended for a workload that needs longer, shortened during an incident.

**Done when**
A key can be issued, used to obtain credentials, and revoked — with revocation taking effect within the stated cache TTL.

**Watch**
- Rotation has no overlap window, so renewal is issue-new → deploy → revoke-old.
- Expired and revoked must return the **same** client-visible outcome. Distinguishing them turns the endpoint into a key-validity oracle.
- Non-expiring keys depend on the proxy retaining every signing key it has ever used for verification (see ADR-013, "Signing Key and Rotation"). If verification keys are ever pruned on a current-plus-previous schedule, keys older than two rotations break silently regardless of their record.
- Auth failures should return one client-visible outcome plus a request id — distinguishing expired from revoked from unknown turns this into a key-validity oracle. That defence needs a matching latency floor, since rejecting locally is measurably faster than a lookup.

See ADR-013.

Depends on source-cooperative/source.coop#546.

---

Part of source-cooperative/source.coop#491.

Issue #232, opened 2026-03-18
[Proposed Feature] UX - Clearer "My Organizations" List


PowerChell:
**Description of Feature:** Add a "My Organizations" tab to the user drop-down (it's fine if it just brings them to their profile where this is listed). A user just got confused on how to find his organizations since they are listed on the profile, but not in their drop-down.

**What value is this feature adding to Source Cooperative?** Easier usability, clearer navigation. 

At large, we should probably be doing a UX revamp, but just noting this since a user just gave this specific feedback.

m-mohr:
Same for the products.

kylebd99:
+1 to this!

PR #570, opened 2026-09-22
feat(accounts): issue, revoke and expire API keys for service accounts

alukach:
**⛔ Blocked by #567** (stacked on its branch) **and by source-cooperative/data.source.coop#233** (the proxy PR for source-cooperative/data.source.coop#231): issuing a key calls the proxy's mint endpoint, which that PR adds, and exchanging one needs the proxy to trust its own issuer and consult the exchanges route here. The `source.coop` half of #548. Part of #491.

## What

Rebuilt on the trust model (#566) in one commit; the earlier history is in this PR's comments.

API keys for environments without OIDC — a server, a scheduler, an instrument — per ADR-013 as amended by ADR-014 (source-cooperative/data.source.coop#232).

**The key** is a JWT the data proxy signs, `sub` the service account, `jti` the record's id, `type: api_key`, `exp` optional. It is handed out as `sck_` + the JWT; the proxy strips the prefix. **That prefix is a decision I've made, flagged for you:** a JWT always begins `eyJ`, so every key matches `sck_eyJ[\w-]+\.[\w-]+\.[\w-]+`, which is what gets registered with secret scanners (#561). This app never stores the key — only its record.

**Issuing** (`issueApiKey`): whoever manages the service account gives a label and an expiry (30/90/365 days, or never). The action writes the record, asks the proxy to sign (`mintApiKey`, calling `POST /.keys` as the issuer through the same Ory ID-token flow `getProxyCredentials` uses), and returns the key once. If the proxy won't sign, the record is removed: no row may look like a live credential. **Revoking** sets `revoked_at`; **expiry** can be changed after issuance, including to never — the decisions from the epic rewrite.

**Exchanging**: `POST /api/v1/service-account-keys/{jti}/exchanges` is what the proxy calls at `/.sts` when a key is presented, authenticated as the service account. It records last use and answers `active` — not revoked, not expired — so revocation takes effect for new exchanges within the proxy's cache TTL (60s, per ADR-013). Credentials already issued live to their session cap, as the epic states.

**Resolving the subject**: `authenticateWithOidcToken` tries `fetchByOryId`, then a service account by id (moved to #566, which needs it for the trust check), so a token whose subject is a service account's id — which is what the proxy forwards after an API-key exchange — resolves to that account. No trust or binding is written for a key: its subject *is* the account. A disabled service account is refused a key. No service account's id can be someone's Ory identity id: #567 namespaces it as `{owner}--{id}`, and a UUID never contains `--`.

**UI**: the service account's page (#567's `ServiceAccountDetail`) gains an API keys section: a row per key with its label, issued / expires / last used, a marker once it is revoked or expired, and Revoke. The section header carries an `IssueApiKeyDialog`, which shows the key once with the copy affordance and the warning. The list row counts live keys as a way to sign in, alongside trusted workflows. The page fetches the keys, and the key actions revalidate it along with the list. Storybook: the list and detail stories carry keys; the dialog has its own story (the mocked action resolves as signed, so submitting reaches the show-once view).

The Nightly Sync page: Signs in as with two trusted workflows; API keys with an Issue an API key button, HPC cron job issued 3/11/2026 expiring 3/11/2027 last used 3/19/2026 with Revoke, and Old laptop marked REVOKED; Can reach; Danger zone

Also: `managedServiceAccount` moves into `src/lib/accounts/service-accounts.ts` so both action modules share one answer to "may this session change this service account".

## Stories

- `ServiceAccountDetail` › **Default** — https://source-coop-ui-git-feat-service-account-keys-radiantearth.vercel.app/?path=/story/features-service-accounts-serviceaccountdetail--default
- `ServiceAccountList` › **Default** — https://source-coop-ui-git-feat-service-account-keys-radiantearth.vercel.app/?path=/story/features-service-accounts-serviceaccountlist--default
- `IssueApiKeyDialog` › **Default** — https://source-coop-ui-git-feat-service-account-keys-radiantearth.vercel.app/?path=/story/features-service-accounts-issueapikeydialog--default

## Testing

- `npx jest` — all suites pass, re-run after the rebase onto #567's namespaced ids. New: `service-account-keys.test.ts` (issue records, signs and returns once; no-expiry keys; a refused signature removes the record; non-manager / bad label / bad expiry refused before any write; revoke only own keys; expiry changes including to never), the exchanges route test (active with last-use recorded; revoked and expired inactive without recording; 401 for another account; 404), and an `oidc.test.ts` case for a proxy-bound subject.
- `npm run type-check`, `next lint`, `npm run build-storybook` — clean.

## Docs and ADRs

data.source.coop: ADR-013 as amended by ADR-014 (#232) is what this implements; the `sck_` prefix and the exchanges route are the two things the proxy side (source-cooperative/data.source.coop#233) matches; that PR also amends ADR-013 to give keys an `aud` claim, which the proxy sets when it signs, so nothing here changes. docs.source.coop: the unattended-workflow guide (source-cooperative/docs.source.coop#34) will carry the `AWS_WEB_IDENTITY_TOKEN_FILE` setup the dialog hints at; nothing existing describes keys.


alukach:
Pushed f93dcac7 for the review finding: the keys table now sets one field at a time with an `UpdateCommand` conditioned on the record existing (`set(jti, field, value)`), and the exchanges route, revoke and expiry each write only the field they own. Title and description still hold.

alukach:
Pushed a follow-up so the comparisons this branch added ask `isServiceAccount` (the guard now lives in `src/types/account.ts`, from #563). Description still holds.

alukach:
Stack #571 rebased to carry #565's no-backfill change (Ory identities stay on `identity_id`; the resolver's Ory-first chain is unchanged), plus one guard from the review: `issueApiKey` refuses a service account whose id is also someone's Ory identity id, since the proxy forwards the subject bare and the person would win. Description updated.

alukach:
Applied the review outcome: no binding is written under the proxy's issuer any more. The resolver tries a person by Ory identity, then a service account by id (never a person's or organization's handle), and `issueApiKey` only writes the key record. This also removes the case where "Remove" on the integrations list would have deleted the binding and silently broken every key. ADR-014 (source-cooperative/data.source.coop#232) amended to match; description updated.

alukach:
Rewritten on the trust model: an `account-trusts` table keyed by the account replaces identity bindings, proof of control is gone, and a workflow names the account in `AWS_ROLE_ARN` (`arn:aws:iam::<service-account-id>:role/FullAccess`). Stack #571 is rebuilt as `main ← #566 ← #567 ← #570`, one commit each; descriptions rewritten. ADR-014 (source-cooperative/data.source.coop#232) amended to match.

alukach:
Rebased onto #567's redesign. The API keys section moved from the old service-account card to the account's own page, `ServiceAccountDetail`. The list row counts live keys, and the key actions revalidate the page. Description and screenshot updated.

alukach:
Rebased onto #567's namespaced ids. New commit 009e255d drops the refusal of a key for an account whose id is also an Ory identity id: a namespaced id always contains `--` and a UUID never does. The keys tests and stories use namespaced ids. Description updated.

alukach:
Rebased onto `main` after #567's squash merge and force-pushed: the branch is now this PR's two commits only. The content is unchanged; the tree matches the previous head exactly. Description updated: #567 no longer blocks this, source-cooperative/data.source.coop#233 still does.

